PR #4, opened 2019-06-06
WAZO-847 HA: fix status code 500 on null remote_address

sduthil:


fblackburn1:
Since it fix a bug, I would prefer to track the change/bug on JIRA. Can you create an issue?

sduthil:
Done: [WAZO-847]

PR #17, opened 2019-08-01
Tests remove yield

sileht:


sileht:
I may have an idea to convert them in pytest fixtures finally.

fblackburn1:
does PR is still valid or we can close it?

sileht:
Closing for now I will repropose or later

PR #24, opened 2019-09-04
Minor documentation fixes for queue options

chartrandf:


fblackburn1:
Do you have any problem with underscore version?
For a (big) historical reason, the right syntax of the API is underscore, but the hyphen version will work too

PR #35, opened 2019-10-16
registrars: fix search test

bloom1:
reason: main_host is autogenerated and the test failed once because we
were searching for `2.3` and the main_host of the HiddenRegistrar was
`55.22.37.54`, thus being returned. Fixed by searching for `1.2.3`
instead.


fblackburn1:
regate

fredericlepied:
regate

PR #36, opened 2019-10-18
WAZO-999: add call_pickup_targets in schema

bloom1:
reason: this is to make it possible to see which users and groups a user
can intercept.
Depends-On: wazo-platform/xivo-dao#14

### DIFF
--- a/integration_tests/suite/base/test_call_pickup_target_group.py
+++ b/integration_tests/suite/base/test_call_pickup_target_group.py
@@ -116,11 +116,8 @@ def test_associate_multi_tenant(
 @fixtures.group()
 @fixtures.group()
 def test_get_user_interceptor_user_relation(call_pickup, user1, user2, user3, line1, line2, line3, group1, group2):
-    with a.user_line(user1, line1), a.user_line(user2, line2), a.user_line(user3, line3):
-        response = confd.groups(group1['id']).members.users.put(users=[user2])
-        response.assert_updated()
-        response = confd.groups(group2['id']).members.users.put(users=[user3])
-        response.assert_updated()
+    with a.user_line(user1, line1), a.user_line(user2, line2), a.user_line(user3, line3), \
+            a.group_member_user(group1, user2), a.group_member_user(group2, user3):
 
         with a.call_pickup_interceptor_user(call_pickup, user1):
             with a.call_pickup_target_group(call_pickup, group1, group2):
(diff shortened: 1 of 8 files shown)


integration_tests/suite/base/test_call_pickup_target_group.py
```diff
@@ -106,6 +106,34 @@ def test_associate_multi_tenant(
     response.assert_match(400, e.different_tenant())
 
 
+@fixtures.call_pickup()
+@fixtures.user()
+@fixtures.group()
+@fixtures.group()
+def test_interceptor_user_has_targets(call_pickup, user1, group1, group2):
```
fblackburn1: don't add number suffix if there are only one resource
```suggestion
def test_interceptor_user_has_targets(call_pickup, user, group1, group2):
```

integration_tests/suite/base/test_call_pickup_target_user.py
```diff
@@ -104,6 +104,34 @@ def test_associate_multi_tenant(main_call_pickup, sub_call_pickup, main_user, su
     response.assert_match(400, e.different_tenant())
 
 
+@fixtures.call_pickup()
+@fixtures.user()
+@fixtures.user()
+@fixtures.user()
+def test_interceptor_user_has_targets(call_pickup, user1, user2, user3):
```
fblackburn1: Use the same semantic as other tests
```suggestion
def test_get_interceptor_user_relation(call_pickup, user1, user2, user3):
```

integration_tests/suite/base/test_call_pickup_target_user.py
```diff
@@ -104,6 +104,34 @@ def test_associate_multi_tenant(main_call_pickup, sub_call_pickup, main_user, su
...
+    response = confd.callpickups(call_pickup['id']).targets.users.put(
+        users=[user2, user3]
+    )
```
fblackburn1: Use `with` statement and association helper

wazo_confd/plugins/user/api.yml
```diff
@@ -399,6 +399,7 @@ definitions:
     - $ref: '#/definitions/UserRelationSwitchboards'
     - $ref: '#/definitions/UserRelationVoicemail'
     - $ref: '#/definitions/UserRelationQueues'
+    - $ref: '#/definitions/UserCallPickupTargets'
```
fblackburn1: Use standard name
```suggestion
    - $ref: '#/definitions/UserRelationCallPickupTargets'
```

wazo_confd/plugins/user/schema.py
```diff
@@ -89,6 +98,33 @@ class UserSchema(BaseSchema):
...
+            set(self._flatten(data.call_pickup_user_targets))
+        )
+
+        data.call_pickup_group_targets_flat = list(
+            set(self._flatten(data.call_pickup_group_targets))
+        )
+        return data
+
+    @post_dump
+    def postdump(self, data):
```
fblackburn1: use better name, like `flatten_call_pickup_targets`

wazo_confd/plugins/user/schema.py
```diff
@@ -89,6 +98,33 @@ class UserSchema(BaseSchema):
         'QueueSchema', only=['id', 'name', 'label', 'links'], many=True, dump_only=True
     )
 
+    @pre_dump
+    def predump(self, data, **kwargs):
```
fblackburn1: use better name

PR #40, opened 2019-11-08
WAZO-1106: add + in regex

fouille:
Correct regex for implement specific username / provider 
Tested on production


bloom1:
JIRA Ticket: WAZO-1106

PR #41, opened 2019-11-12
trunk: add the endpoint sip name to the relation

pc-m:


pc-m:
> But if you do not add `name` to the line resource too, the documentation is wrong

Eagle eye Blackburn strikes again.

Fixed thanks

PR #46, opened 2019-12-04
use only ipv4 nameservers

lebauce:


lebauce:
A simplier approach could also be to only select IPv4 nameservers:
```
diff --git a/wazo_confd/plugins/wizard/service.py b/wazo_confd/plugins/wizard/service.py
index aee9f95..eab25a8 100644
--- a/wazo_confd/plugins/wizard/service.py
+++ b/wazo_confd/plugins/wizard/service.py
@@ -18,7 +18,7 @@ from .notifier import build_notifier
 from .validator import build_validator
 
 USERNAME_VALUES = '2346789bcdfghjkmnpqrtvwxyzBCDFGHJKLMNPQRTVWXYZ'
-NAMESERVER_REGEX = r'^nameserver (.*)'
+NAMESERVER_REGEX = r'^nameserver (\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})$'
 DEFAULT_ADMIN_POLICY = 'wazo_default_admin_policy'
 ASTERISK_AUTOPROV_CONFIG_FILENAME = '/etc/asterisk/pjsip.d/05-autoprov-wizard.conf'
 ASTERISK_AUTOPROV_CONFIG_TPL = '''\
```

sboily:
We never check is we support ipv6.

fblackburn1:
:laughing: are you really sure you want to quick doing this thing? :stuck_out_tongue: 
If yes, so you need to check [xivo-sysconfd](https://github.com/wazo-platform/xivo-sysconfd/blob/master/xivo_sysconf/modules/resolvconf.py#L132)

lebauce:
@sboily @fblackburn1 Do you prefer the workaround I posted in my first comment ?

fblackburn1:
> @sboily @fblackburn1 Do you prefer the workaround I posted in my first comment ?

It's your, depending how much time you want to test it and fix it.
In all case, this PR alone is not enough

lebauce:
@fblackburn1 This fix was more about not crashing because /etc/resolv.conf contains IPv6 addresses :-) And yes, for sure we need to a lot more to properly support IPv6

lebauce:
@fblackburn1 Thanks

fredericlepied:
regate

PR #47, opened 2019-12-05
tests: reproduced non-linked destination issue

sileht:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/21

### DIFF
--- a/integration_tests/suite/base/test_incalls.py
+++ b/integration_tests/suite/base/test_incalls.py
@@ -5,6 +5,7 @@ from hamcrest import (
     all_of,
     assert_that,
     contains_inanyorder,
+    equal_to,
     empty,
     has_entries,
     has_entry,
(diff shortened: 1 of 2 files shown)


sileht:
recheck

fblackburn1:
In general if you want to do this kind of test, use the resource in xivo-dao
Ex:
https://github.com/wazo-platform/xivo-dao/blob/master/xivo_dao/resources/ivr/tests/test_dao.py#L295

sileht:
> This fix as nothing to do with wazo-confd, it's purely a database coherence

The bug was report with a chain of API calls to do, so I write the same chain in tests to understand what happen.

Then, I wrote another minimal test in xivo-dao that trigger the bug too: https://github.com/wazo-platform/xivo-dao/pull/22/files#diff-50d0dd21dde6c97ee3f912187f78c412R108

I will move it in xivo_dao/resources/incall/tests/test_dao.py as your recommend tomorrow.

We can close that one if you thing this new tests does worse to be kept

fblackburn1:
> Then, I wrote another minimal test in xivo-dao that trigger the bug too: https://github.com/wazo-platform/xivo-dao/pull/22/files#diff-50d0dd21dde6c97ee3f912187f78c412R108
> 
> I will move it in xivo_dao/resources/incall/tests/test_dao.py as your recommend tomorrow.
> 
> We can close that one if you thing this new tests does worse to be kept

No need to move the test, since the logic is in the sqlalchemy schema, its totally fine

PR #48, opened 2019-12-09
confbridge: add links to the sample config file in the API spec

pc-m:
Add links to the sample configuration file for confbridge in the API spec since
the help in the wiki misses some of the fields that are available.


fredericlepied:
recheck

bloom1:
recheck

PR #53, opened 2019-12-18
Pull request for WAZO-1436-add-missing-joinedload

fblackburn1:
refactor db test to be used by other resources
test the number of db requests for list users


fblackburn1:
recheck

PR #54, opened 2019-12-18
assert that GET /extensions only trigger one database query

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/26


pc-m:
recheck

PR #55, opened 2019-12-26
line endpoints: publish events on association and dissociation

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-bus/pull/14

This PR can be merged independently from the calld part of this feature so I'm adding the mergeit tag on bus and confd now in case someone would like to review those parts first.


fblackburn1:
recheck

PR #58, opened 2020-01-09
user: do not execute pre_dump for field that are not required

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/29


sboily:
:hearts:

fblackburn1:
> hearts

Don't be too excited, it was a regression from new code added in this iteration

sboily:
> > hearts
> 
> Don't be too excited, it was a regression from new code added in this iteration

:broken_heart:

fblackburn1:
recheck

PR #63, opened 2020-01-28
Pull request for WAZO-1452-ha-update-sip-provisioning

sduthil:
registrars: do not return 404 if a device is not found
HA: update SIP and provisioning config

### DIFF
--- a/wazo_confd/plugins/ha/service.py
+++ b/wazo_confd/plugins/ha/service.py
@@ -16,18 +16,14 @@ class HAService:
 
     def edit(self, form):
         if form['node_type'] == 'disabled':
-            self._update_sip_general_options({
-                'minexpiry': '60',
-                'maxexpiry': '3600',
-                'defaultexpiry': '120',
-            })
+            self._update_sip_general_options(
+                {'minexpiry': '60', 'maxexpiry': '3600', 'defaultexpiry': '120'}
+            )
             self._update_provisioning_options(remote_address=None)
         elif form['node_type'] == 'master':
-            self._update_sip_general_options({
-                'minexpiry': '180',
-                'maxexpiry': '300',
-                'defaultexpiry': '240',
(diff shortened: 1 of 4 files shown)


wazo_confd/plugins/ha/service.py
```diff
@@ -1,14 +1,43 @@
...
+        self.sip_general_service = sip_general_service
         self.sysconfd = sysconfd
 
     def get(self):
         return self.sysconfd.get_ha_config()
 
     def edit(self, form):
+        if form['node_type'] == 'disabled':
+            self._update_sip_general_options(
+                {'minexpiry': '60', 'maxexpiry': '3600', 'defaultexpiry': '120'}
```
bloom1: Is it common for an admin to set different values for these options?

wazo_confd/plugins/registrar/service.py
```diff
@@ -29,7 +34,13 @@ def edit(self, registrar, updated_fields=None):
         # Update all affected devices
         for line in lines:
             if line.device_id not in devices_updated:
-                self.device_updater.update_for_line(line)
+                try:
+                    self.device_updater.update_for_line(line)
+                except NotFoundError:
+                    logger.error(
+                        f'Could not update device "{line.device_id}": device not found'
```
bloom1: f-strings! Nice! :smiley_cat:

sduthil: Actually, you made me realize that we should not use them for logging, because it forces eager evaluation, instead of lazy evaluation with
`logger.error('Could not update device "%s", line.device_id)`

wazo_confd/plugins/ha/service.py
sduthil: Maybe, but they're necessary for high-availability to function correctly, so we just overwrite them.

PR #64, opened 2020-01-30
test: check incall destination when it's deleted

fblackburn1:
reason: for avoid breaking change we do not authorize to have
destination=null. So a fix has been added in xivo-dao to always return
destination={type: none} and we ensure that tests pass here

Depends-on: https://github.com/wazo-platform/xivo-dao/pull/34


fblackburn1:
recheck

PR #67, opened 2020-02-06
RFC: PJSIP: add API spec

pc-m:
This PR is not meant to be merged, I want to be able to share the API spec and get comments on it.

I've created a gist [here](https://gist.github.com/pc-m/a6d7a1971ce2fb45b26fca45e90d08e0) if you want to copy/paste the spec into [swagger](https://editor.swagger.io/) without having to setup a dev environment to render it.

Add comments here and I will keep the gist updated if I make changes to the spec.

# Overview

The API is spread between 2 parts, `/endpoints/sip` and `/asterisk/sip`. The part in `/asterisk/sip` is for shared configuration. It includes the `global`, `system` sections as well as a CRUD for transports. The `/endpoints/sip` section is used to create the endoint/aor/auth/registration sections. Each section can have parents which will be used as "templates".

The template system will allow each tenant to have one or many sip general configuration. For example `sip` and `webrtc`. Then each line/trunk for this tenant would select a parent and only specify fields that need to be customized.

To allow the UI to be helpful to the administrator I have included a resource to fetch all configuration options, with there name, synopsis, description and default values. This resource is available under `/asterisk/sip/doc`

# Resource creation

## Global settings

Global settings should be created during the wizard or in the populate.sql script. This include

* transport-udp
* transport-wss
* transport-tls?  <-- We would need to generate keys to do that
* global section
* system section

## Tenant related templates

At the creation of each tenant the following configuration would have to be created

* A "normal" SIP config for lines and trunks
* A webrtc config for webrtc lines

When creating a "concrete" endpoint_sip only the username and secret would be necessary or auto-generated. All other fields could be inherited from the template.

# What's different from other resources

I've tried to create an API that is consumer friendly instead of being a mapping of the DB. What this means is that some of the DB relationships are not mapped in the resources.

I did not create an association resource between the transports and the endpoint_sip. The transport can be specified in the body of the post or put and the appropriate relationship will be created. I was thinking of allowing the name of the transport to be accepted in the body of the request.

Deleting transports that are still associated will required a param to indicate which transport to use for the associated endpoints. It won't be possible to delete the last remaining transport.


fblackburn1:
The second part of the url for endpoints starting with `/asterisk/` is the name of the config file. In this case it should be `/asterisk/pjsip`

The `parent_uuids`, why is it not by section and it is global to the `endpoints/sip` object ? To avoid conflict with potential asterisk option `parent_uuid`?

> When creating a "concrete" endpoint_sip only the username and secret would be necessary or auto-generated. All other fields could be inherited from the template.

How to determine the "concrete" state? Today, it cause some UI issues when we auto-generate username/password on endpoint that don't need it

> I've tried to create an API that is consumer friendly instead of being a mapping of the DB. What this means is that some of the DB relationships are not mapped in the resources.

Can you give us an example?
Do you talk about adding line or trunk relation in GET as readonly param?
Or you talk about `parent_uuids` is not `parents: [{uuid: ...}, {uuid, ...}]`?
For the last one I can see some issues, to do not have the name of the parent for the UI if we do only a `parent_uuids`.

> I did not create an association resource between the transports and the endpoint_sip. The transport can be specified in the body of the post or put and the appropriate relationship will be created. 

I don't see which key are you talking to associate, can you give me an example?

> I was thinking of allowing the name of the transport to be accepted in the body of the request.

In this case you will enforce the uniqueness of the name? Or we should force the name prefix according to the tenant?

> Deleting transports that are still associated will required a param to indicate which transport to use for the associated endpoints. It won't be possible to delete the last remaining transport.

I like the idea, never thought about this problematic.
 I'm curious to listen other possibilities (if you have one).

You confirm the typo from your post: `/endpoints/sip` not `/endpoint_sip` ?

Thanks for sharing :+1:

pc-m:
> The second part of the url for endpoints starting with `/asterisk/` is the name of the config file. In this case it should be `/asterisk/pjsip`

I re-used to old routes but I was not sure it was a good idea to use the `/asterisk` URL since we are not doing a direct mapping of the configuration.

> 
> The `parent_uuids`, why is it not by section and it is global to the `endpoints/sip` object ? To avoid conflict with potential asterisk option `parent_uuid`?

I wanted to expose expose the whole configuration as a coherent set to avoid the burden of choosing multiple templates when configuring an endpoint

For example, if I create a new webrtc line I choose the webrtc template from my tenant. Having parent for each section I would choose the webrtc_endpoint, webrtc_aor, webrtc_auth and for a "normal" line I would choose line_endpoint, line_aor, line_aor.

> 
> > When creating a "concrete" endpoint_sip only the username and secret would be necessary or auto-generated. All other fields could be inherited from the template.
> 
> How to determine the "concrete" state? Today, it cause some UI issues when we auto-generate username/password on endpoint that don't need it

The `template` boolean field of the endpoint_sip section indicates if it's a template or not

> 
> > I've tried to create an API that is consumer friendly instead of being a mapping of the DB. What this means is that some of the DB relationships are not mapped in the resources.
> 
> Can you give us an example?
> Do you talk about adding line or trunk relation in GET as readonly param?
> Or you talk about `parent_uuids` is not `parents: [{uuid: ...}, {uuid, ...}]`?

What I meant here is that each section will most likely end in different table but the user of the API will do a single POST to create a config not one for it's AOR, endpoint, auth, registration, outbound registration.

> For the last one I can see some issues, to do not have the name of the parent for the UI if we do only a `parent_uuids`.

Good point, I will use a dict
> 
> > I did not create an association resource between the transports and the endpoint_sip. The transport can be specified in the body of the post or put and the appropriate relationship will be created.
> 
> I don't see which key are you talking to associate, can you give me an example?

1. We could also accept the name or uuid of the transport to make the body more similar for the whole body. And confd would to the job of finding the transport.

`{"endpoint": {"transport": "transport-udp"}}`

2. The API does not specify that at the moment but that's what I add in mind.

`{"endpoint": {"transport": "{"uuid": "<uuid>", "name": "transport-udp"}}}`

I have a preference for 1 in that context

`transport-udp` is a transport created by the `/asterisk/sip/transports` CRUD.

> 
> > I was thinking of allowing the name of the transport to be accepted in the body of the request.
> 
> In this case you will enforce the uniqueness of the name? Or we should force the name prefix according to the tenant?

Transports are global to the engine and there name MUST to be unique.

Asterisk refuses to have 2 transports of the same protocol binding on the same port, making it impossible to create a transport for each tenant.

> 
> > Deleting transports that are still associated will required a param to indicate which transport to use for the associated endpoints. It won't be possible to delete the last remaining transport.
> 
> I like the idea, never thought about this problematic.
> I'm curious to listen other possibilities (if you have one).

1. My first idea was to have a column default on the transport (only one of them should be default) and assign to the default on delete.

2. block delete if associated, I hate this one

> 
> You confirm the typo from your post: `/endpoints/sip` not `/endpoint_sip` ?

fixing it
> 
> Thanks for sharing

fblackburn1:
I finally read the whole answer, sorry for the delay :sweat_smile: 
I'm ok :+1: with all your answers

> > > Deleting transports that are still associated will required a param to indicate which transport to use for the associated endpoints. It won't be possible to delete the last remaining transport.
> > 
> > I like the idea, never thought about this problematic.
> > I'm curious to listen other possibilities (if you have one).
>
> 1. My first idea was to have a column default on the transport (only one of them should be default) and assign to the default on delete.
> 

Yeah the main disadvantage with this solution is with permission. If the default transport is set on the /sip/transport, only the owner of the engine can change the default one. Because I guess that endpoint should be not available to tenant. But if we can decide the default on deletion, it solve this issue.

Another idea is to have en endpoint of `global configuration by tenant` to specify this kind of configuration ... but well is not incompatible with your solution
> 2. block delete if associated, I hate this one
> 
I hate too

sduthil:
Thank you for this nice description!

My questions:
* What about API retrocompatibility? `/endpoints/sip` will be completely different from what it used to be, right?
* Why does an endpoint have multiple parents? Which one takes priority if parent templates conflict? Can a parent be a non-template endpoint? If not, can we name the key `templates`?
* About transport deletion, which solution is preferred? 1. Using a `default` key in the body of the transports or 2. using a `fallback` query-string argument to `DELETE /sip/transports`? The OpenAPI spec describes 1. but the discussion mentions 2. BTW, `default` key lets me think that this transport will be used when *creating* an endpoint without transport, not when *deleting* a transport. So I'd prefer a `fallback` key in the body instead of `default`.

pc-m:
> Thank you for this nice description!
> 
> My questions:
> 
> * What about API retrocompatibility? `/endpoints/sip` will be completely different from what it used to be, right?

Yes, I reused the same resource and did not care about retro-compatibility. I could use `/endpoints/pjsip` instead and add a translation to `/endpoints/sip` to keep the API working while removing the underlying DB.

The discussion about API version change has been completely overlooked in this proposal.

> * Why does an endpoint have multiple parents? Which one takes priority if parent templates conflict? Can a parent be a non-template endpoint? If not, can we name the key `templates`?

Given a template general_sip_line
Given a template general_webrtc_line
Given a template user_preference_pascal

Line 1 endpoints/sip
parents: user_preference_pascal, general_sip_line
name: foobar
password: iddqd

Line 2 endpoints/sip
parents: user_preference_pascal,general_webrtc_line
name: mywebrtc
password: idbehold

In these 2 cases the user_preference_pascal has the highest priority and then the "general" templates apply. The "general" templates are global to the current tenant. The user_preference is specific to a given user.

I have no intention of implementing the user preference part right now but I did not want to have to create a user_preference_webrtc_pascal and a user_preference_sip_pascal if/when we do want to implement that.

> * About transport deletion, which solution is preferred? 1. Using a `default` key in the body of the transports or 2. using a `fallback` query-string argument to `DELETE /sip/transports`? The OpenAPI spec describes 1. but the discussion mentions 2. BTW, `default` key lets me think that this transport will be used when _creating_ an endpoint without transport, not when _deleting_ a transport. So I'd prefer a `fallback` key in the body instead of `default`.

I prefer the fallback query-string and I think it's also François' choice, so I guess it's going to be `fallback`

sduthil:
>> What about API retrocompatibility? /endpoints/sip will be completely different from what it used to be, right?

> Yes, I reused the same resource and did not care about retro-compatibility. I could use /endpoints/pjsip instead and add a translation to /endpoints/sip to keep the API working while removing the underlying DB.
The discussion about API version change has been completely overlooked in this proposal.

Maybe @sboily has an opinion on this? Should we expose the `pjsip` name in the URL of Wazo Platform API, since the body is closely related to Asterisk PJSIP config?

sduthil:
>> What about API retrocompatibility? /endpoints/sip will be completely different from what it used to be, right?

> Yes, I reused the same resource and did not care about retro-compatibility. I could use /endpoints/pjsip instead and add a translation to /endpoints/sip to keep the API working while removing the underlying DB.
The discussion about API version change has been completely overlooked in this proposal.

Maybe @sboily has an opinion on this? Should we expose the `pjsip` name in the URL of Wazo Platform API, since the body is closely related to Asterisk PJSIP config?

sduthil:
>> What about API retrocompatibility? /endpoints/sip will be completely different from what it used to be, right?

> Yes, I reused the same resource and did not care about retro-compatibility. I could use /endpoints/pjsip instead and add a translation to /endpoints/sip to keep the API working while removing the underlying DB.
The discussion about API version change has been completely overlooked in this proposal.

Maybe @sboily has an opinion on this? Should we expose the `pjsip` name in the URL of Wazo Platform API, since the body is closely related to Asterisk PJSIP config?

fblackburn1:
> > * What about API retrocompatibility? `/endpoints/sip` will be completely different from what it used to be, right?
> 
> Yes, I reused the same resource and did not care about retro-compatibility. I could use `/endpoints/pjsip` instead and add a translation to `/endpoints/sip` to keep the API working while removing the underlying DB.

Technically pjsip is not a protocol
IMHO we should not support two different methods to configure an endpoint. It will create overload for us and confusion for user. The benefit to keep two methods is if all the work cannot be done in one sprint otherwise it's a waste of time.

pc-m:
I added a new field on an endpoint_sip called `slave_only` which is a boolean. If this field is true then the configuration will only be applied when configuration a node in slave mode.

This is to allow overriding some of the fields in that configuration that might contain IP addresses.

sduthil:
I would like to avoid introducing `slave` config, as much as possible. I can't think of a better way right now though. Do you have alternatives?

fblackburn1:
disclaimer: I don't really know which options we are talking about
Using a specific table for these configuration and avoid to duplicate this table
But you will need to connect to the slave to configuration these options

pc-m:
I see 2 cases that are possible.

1. some ip are configured and are the same on the master and slave
2. some ip are configured and are NOT the same on the master and slave

I cannot see any option other that having a config template that is applied on the slave only to override the master config.

Not replicating tables does not work here because that would exclude the endpoint_sip table which is needed for lines and trunks.

Not replicating variables for a given section of a given file in the asterisk_* tables seems complicated and too complex for the script we are using at the moment. This also applies to the ice candidates of `rtp.conf` which has a similar problem.

What I like about the `slave` config is that is can be configured from the master and removes logic from the DB synchronisation which should be as simple as it can IMO.
What I dislike is the word `slave` which refers to slavery. But I guess that should be addressed separately.

sduthil:
Another alternative, close to the `slave_only` option, would be to have a `wazo_uuid` or `asterisk_id` option that would make the SIP option only apply on the Wazo Platform having the correct UUID. I don't know if it could have other uses, such as multiple Asterisk working together, with specific SIP options for each instance.

pc-m:
> Another alternative, close to the `slave_only` option, would be to have a `wazo_uuid` or `asterisk_id` option that would make the SIP option only apply on the Wazo Platform having the correct UUID. I don't know if it could have other uses, such as multiple Asterisk working together, with specific SIP options for each instance.

I like that option, I think the `wazo_uuid` are the same on a master/slave cluster but the asterisk id would definitely be a good candidate since it's an Asterisk specific configuration.

fblackburn1:
@pc-m 
PJSIP being merged, Is this PR still relevant or can we close it?

PR #74, opened 2020-02-28
accessd_feature: restart phoned when update values

fblackburn1:
It is a temporary solution. xivo-sysconfd should listen on events and
restart service asynchronous ... or better, phoned should listen on
events and reload service according ... or even better, phoned should
handle its own configuration.

Summary: this is one of the worst solution that can be done

EDIT: but the best for the time involved

### DIFF
--- a/wazo_confd/_sysconfd.py
+++ b/wazo_confd/_sysconfd.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2013-2020 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import requests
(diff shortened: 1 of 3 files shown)


sboily:
Are you sure there is no other solution?

fblackburn1:
> Are you sure there is no other solution?

It's the best solution according the time investment... it's more a solution to make some people happy and create nice statistics ;)
But don't think I'm against this, we have a lot of priorities and this solution is good enough. In all case I created a more [specific ticket](https://wazo-dev.atlassian.net/browse/WAZO-1619) to switch the phoned config to its own database

sboily:
We definitively want to remove this service from wazo, and adding a new line of code inside this service broke my heart... ;). I propose to talk about that before merging this PR this next week.

pc-m:
> We definitively want to remove this service from wazo, and adding a new line of code inside this service broke my heart... ;). I propose to talk about that before merging this PR this next week.

I merged this PR, so now you're heart broken. We can still talk about other solutions to this problem :)

PR #75, opened 2020-03-02
Wazo 1612 add the /asterisk/pjsip/system resource

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-bus/pull/16
Depends-On: https://github.com/wazo-platform/xivo-manage-db/pull/35


fblackburn1:
recheck

PR #76, opened 2020-03-02
user-import: enforce line_protocol when extension

fblackburn1:
reason: otherwise the extension and user are created but not associated
which is useless


fredericlepied:
recheck

wazo_confd/plugins/user_import/creators.py
```diff
@@ -187,6 +188,14 @@ def create(self, fields, tenant_uuid):
         exten = fields.get('exten')
         context = fields.get('context')
         if exten and context:
+            firstname = fields.get('firstname')
```
pc-m: I can't figure out why you tested for the firstname here

fblackburn1: To avoid breaking too much the current implementation
If you want to import only a exten/context without user

fblackburn1: After investigation I cannot remove it. I tried to make code clearer

PR #81, opened 2020-03-19
Add id to user Service and Forward events

bloom1:
Depends-On: https://github.com/wazo-platform/xivo-bus/pull/18


fblackburn1:
recheck

PR #104, opened 2020-05-20
use endpoint_*_id/register_*_id trunk columns

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/45
Depends-on: https://github.com/wazo-platform/xivo-manage-db/pull/43


pc-m:
recheck

PR #105, opened 2020-05-21
Pull request for improve-docker

fblackburn1:
docker: create default directory to base image
reduce docker image size
docker: use buster slim
use python3 syntax for "super"


pc-m:
regate

PR #110, opened 2020-06-03
nginx: increase timeout for ha endpoint

fblackburn1:
I don't know if it's the right choice (add regex) or if we should only set the same timeout for all endpoints. The same logic can also be applied to the `client_max_body_size`


fblackburn1:
I force push and apply timeout for all endpoints

PR #122, opened 2020-07-08
Pull request for WAZO-1856-sync-db

fblackburn1:


wazo_confd/sync_db.py
```diff
@@ -0,0 +1,75 @@
...
+
+    init_db_from_config(config)
+    with session_scope() as session:
+        tenants = session.query(Tenant).all()
+        confd_tenants = set(tenant.uuid for tenant in tenants)
+        logger.debug('wazo-confd tenants: %s', confd_tenants)
+
+        removed_tenants = confd_tenants - auth_tenants
+        for tenant in removed_tenants:
+            logger.info('Removing tenant: %s... (SKIP)', tenant)
```
pc-m, line 64: Why did you log `(SKIP)`?

fblackburn1: Because the tenant is not removed, only some resources in this tenant
But there is probably a better way to explain that than SKIP

pc-m: you're right, I did not notice that the tenant uuid was still in the confd tenant table after the cleanup. Skip makes sense then.

PR #126, opened 2020-07-09
PJSIP Template - Fix api spec for key/value fields

chartrandf:
Fix spec based on what we received form the API


wazo_confd/plugins/endpoint_sip/api.yml
```diff
@@ -237,34 +237,40 @@ definitions:
...
         default: autogenerated
-  SIPSectionOptions:
-    properties:
-      option:
-        type: string
-        description: The name of the PJSIP option as seen in an Asterisk configuration file
-      value:
-        type: string
-        description: The value for this configuration option
+  KeyValueOption:
```
fblackburn1, line 240: Be careful, your key must be unique and it's not the case here. It may be overwritten by already defined KeyValueOption. If KeyValueOption is used by many schemas, it should be moved to api plugin

chartrandf: Ahh, good catch. I thought it was the right way to do it. I took a part of code from theses files 🙈  I'll make a PR to fix that.

PR #128, opened 2020-07-10
Pjsip remove context

pc-m:
this PR removes the context from SIP endpoints.

A usable endpoint will always be associated to a line or a trunk, both have a context.

I also added some fixes that are unrelated, sorry about that.


chartrandf:
👀  following this PR

PR #136, opened 2020-08-03
WAZO-1809: add tenant_uuid to forward and service events

bloom1:
Depends-On: https://github.com/wazo-platform/xivo-bus/pull/23


bloom1:
recheck

fblackburn1:
regate

PR #148, opened 2020-09-10
Wazo 1965 merged pjsip options

pc-m:
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/73

### DIFF
--- a/wazo_confd/plugins/endpoint_sip/schema.py
+++ b/wazo_confd/plugins/endpoint_sip/schema.py
@@ -4,7 +4,7 @@
 import logging
 
 from marshmallow import fields, EXCLUDE, post_dump
-from marshmallow.validate import Length
+from marshmallow.validate import Length, OneOf
 from marshmallow.utils import get_value
 
 from wazo_confd.helpers.mallow import (
@@ -18,10 +18,12 @@ from wazo_confd.helpers.mallow import (
 logger = logging.getLogger(__name__)
 
 options_field = fields.List(
-    PJSIPSectionOption(option_regex=None),
-    missing=[],
-    validate=Length(max=512),
+    PJSIPSectionOption(option_regex=None), missing=[], validate=Length(max=512),
 )
+
+
+class GETQueryStringSchema(BaseSchema):
+    view = fields.String(validate=OneOf(['merged']), missing=None)
 
 
 class EndpointSIPRelationSchema(BaseSchema):
--- a/wazo_confd/plugins/endpoint_sip/schema.py
+++ b/wazo_confd/plugins/endpoint_sip/schema.py
@@ -18,7 +18,9 @@ from wazo_confd.helpers.mallow import (
 logger = logging.getLogger(__name__)
 
 options_field = fields.List(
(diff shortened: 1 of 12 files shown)


wazo_confd/plugins/endpoint_sip/schema.py
```diff
@@ -70,6 +72,60 @@ class EndpointSIPSchema(_BaseSIPSchema):
...
 
 
+class MergedEndpointSIPSchema(BaseSchema):
+    def get_attribute(self, obj, attr, default):
+        only = getattr(self.declared_fields[attr], 'only', None)
+        if attr.endswith('_section_options') and only:
+            options = get_value(obj, attr)
+            return [[key, value] for key, value in options if key in only]
+        return super().get_attribute(obj, attr, default)
+
```
fblackburn1, line 82:
```suggestion
```
This code is not used for now

wazo_confd/plugins/endpoint_sip/schema.py
```diff
@@ -70,6 +72,60 @@ class EndpointSIPSchema(_BaseSIPSchema):
...
+    name = fields.String(validate=PJSIPSection())
+    label = fields.String(validate=Length(max=128), allow_none=True)
+
+    combined_aor_section_options = options_field
+    combined_auth_section_options = options_field
+    combined_endpoint_section_options = options_field
+    combined_identify_section_options = options_field
+    combined_registration_section_options = options_field
+    combined_registration_outbound_auth_section_options = options_field
+    combined_outbound_auth_section_options = options_field
```
fblackburn1, line 94: Any reason to use a prefix instead of no prefix?
IMO if I request a merged view, I don't want to handle other key name. But it's only a preference, not a strong opinion

fblackburn1 (changes requested):
FYI About your comment
```
Mock and @hybrid_property does not mix well
```
I think we can use only `property` because we don't define `expression` and autogenerated `expression` (by sqlalchemy when it's not defined) will not work because our getter/setter are too complex
(it was only for your information, not to change)

wazo_confd/plugins/endpoint_sip/schema.py
fblackburn1: I think I wrongly read this schema ! instead of doing a post_dump, you can use `attribute`

fblackburn1 (approved):

fblackburn1:
recheck

wazo_confd/plugins/endpoint_sip/schema.py
```diff
@@ -70,6 +81,18 @@ class EndpointSIPSchema(_BaseSIPSchema):
     line = fields.Nested('LineSchema', only=['id', 'links'], dump_only=True)
 
 
+class MergedEndpointSIPSchema(EndpointSIPSchema):
```
pc-m, line 84: I love that version :+1:

But you removed the option merging when you removed the post-dump

fblackburn1: thank you, fixed

fblackburn1 (dismissed):

fblackburn1 (approved):

PR #154, opened 2020-09-18
pjsip: rename global-trunk to trunk-with-registration

pc-m:
this template only adds registration options and generate warnings in the CLI
when trunks do not have a registration


fblackburn1:
recheck

fblackburn1 (approved):

PR #170, opened 2020-11-16
use flask helpers to avoid duplicate requests

fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-lib-python/pull/60

reason: when using recurse=True, flask helpers already fetch tenants. So
instead to fetch list twice, we use the same list


sduthil (approved):

sduthil:
recheck

PR #180, opened 2021-01-11
trunk endpoint sip: add the client_uri to registration section

pc-m:
the client_uri field is required to match the endpoint to a trunk registration
event.


fblackburn1 (changes requested):
event when trunk_endpoint sip are associated should have this info, no?

pc-m:
> event when trunk_endpoint sip are associated should have this info, no?
> ref: https://github.com/wazo-platform/wazo-confd/blob/master/wazo_confd/plugins/trunk_endpoint/notifier.py#L32

 Yes, thanks. I'll add that

fblackburn1 (approved):

PR #187, opened 2021-03-05
moh: deprecate mp3

fblackburn1:
reason: this notice was already in the wazo documentation
(wazo-platform.org)


fblackburn1:
recheck

PR #193, opened 2021-03-22
Nes 906 group name

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/121

### DIFF
--- a/integration_tests/suite/base/test_event_handlers.py
+++ b/integration_tests/suite/base/test_event_handlers.py
@@ -1,4 +1,4 @@
-# Copyright 2020 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2020-2021 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from hamcrest import (
@@ -6,6 +6,7 @@ from hamcrest import (
     contains_inanyorder,
     has_entries,
     empty,
+    starts_with,
 )
 
 from xivo_test_helpers import until
@@ -20,7 +21,7 @@ def test_create_default_templates_when_not_exist():
     assert_that(response.items, empty())
 
     with BaseIntegrationTest.create_auth_tenant(CREATED_TENANT):
-        BusClientHeaders.send_tenant_created(CREATED_TENANT)
+        BusClientHeaders.send_tenant_created(CREATED_TENANT, 'myslug')
 
     def templates_created():
         response = confd.endpoints.sip.templates.get(wazo_tenant=CREATED_TENANT)
@@ -35,4 +36,10 @@ def test_create_default_templates_when_not_exist():
             ),
         )
 
+    def slug_created():
+        # There's no API to check the tenant slug. It is part of any new group name
+        response = confd.groups.post({'label': 'ignore'}, wazo_tenant=CREATED_TENANT)
+        assert_that(response.item, has_entries(name=starts_with('grp-myslug-')))
+
     until.assert_(templates_created, tries=5)
+    until.assert_(slug_created, tries=5)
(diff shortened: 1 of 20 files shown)


integration_tests/suite/base/test_event_handlers.py
```diff
@@ -35,4 +36,9 @@ def templates_created():
             ),
         )
 
+    def slut_created():
+        # There's no API to check the tenant slug. It is part of any new group name
+        group = confd.groups.create({'label': 'ignore'}, wazo_tenant=CREATED_TENANT)
+        assert_that(group, has_entries(name=starts_with('grp-myslug-')))
+
     until.assert_(templates_created, tries=5)
```
fblackburn1, line 44:
```suggestion
    def slut_created():
        # There's no API to check the tenant slug. It is part of any new group name
        group = confd.groups.create({'label': 'ignore'}, wazo_tenant=CREATED_TENANT)
        assert_that(group, has_entries(name=starts_with('grp-myslug-')))

    until.assert_(templates_created, tries=5)
    until.assert_(slug_created, tries=5)
```

wazo_confd/plugins/group/validator.py
```diff
@@ -18,20 +14,12 @@ class GroupValidator(Validator):
     def validate(self, group):
         self.validate_unique_name_through_queue(group)
 
+    # TODO(pc-m): remove this validation once queues get there name generated
     def validate_unique_name_through_queue(self, group):
         queue = queue_dao.find_by(name=group.name)
         if queue is not None:
             raise errors.resource_exists('Queue', name=group.name)
```
fblackburn1, line 21: Can we remove this?

wazo_confd/sync_db.py
```diff
@@ -87,6 +92,13 @@ def main():
             tenant = tenant_dao.find_or_create_tenant(tenant_uuid)
             default_sip_template_service.generate_sip_templates(tenant)
 
+        for tenant_uuid in confd_tenant_without_slugs:
+            slug = auth_tenant_slugs.get(tenant_uuid)
+            if not slug:
+                continue
+            tenant = tenant_dao.find_or_create_tenant(tenant_uuid)
+            tenant.slug = slug
```
fblackburn1, line 100: by good practice we should flush session after this operation (with flush directly or wrap it in `tenant_dao.update(tenant)`)

integration_tests/suite/base/test_event_handlers.py
pc-m: Done, I also fixed the `slug_created` typo

wazo_confd/plugins/group/validator.py
pc-m: I'd rather wait until we do the same thing with queues and avoid someone messing around with queue name and end up doing something that breaks everything and then open a bug.

wazo_confd/helpers/resource.py
```diff
@@ -39,3 +39,6 @@ def delete(self, resource):
         self.validator.validate_delete(resource)
         self.dao.delete(resource)
         self.notifier.deleted(resource)
+
+    def update_form(self, form):
+        return form
```
fblackburn1, line 44: This layer should receive only object (sqlalchemy) or id. The form transformation is done in the restful layer

fblackburn1 (changes requested):

wazo_confd/plugins/group/validator.py
fblackburn1: Ok :+1:

wazo_confd/sync_db.py
pc-m: which does not exist :sweat_smile:

fblackburn1:
recheck

wazo_confd/sync_db.py
fblackburn1: Hahaha

fblackburn1 (approved):

fblackburn1 (dismissed):

PR #202, opened 2021-06-15
Wazo 2314 user moh on incall

pc-m:


wazo_confd/helpers/destination.py
```diff
@@ -417,6 +417,34 @@ def make_user_fields_flat(self, data):
...
+
+        return data
+
+    @post_load
+    def merge_action(self, data):
+        ring_time = data.pop('ring_time', None)
+        moh_uuid = data.pop('moh_uuid', None)
+
+        actionarg2 = ''
+        if ring_time:
```
bloom1, line 440: What if `ring_time` is 0? Is that a use case?

pc-m: You are right, fixed!

wazo_confd/plugins/schedule/validator.py
```diff
@@ -1,8 +1,107 @@
...
+
+        return data
+
+    @post_load
+    def merge_action(self, data):
+        ring_time = data.pop('ring_time', None)
+        moh_uuid = data.pop('moh_uuid', None)
+
+        fallback_actionargs = ''
+        if ring_time:
```
bloom1, line 55: Same thing here :smile:

pc-m: :see_no_evil:

bloom1 (approved):

PR #203, opened 2021-06-16
switchboard: fix nulled moh on missing fields

pc-m:


fblackburn1 (changes requested):
now it's impossible to set moh to null

fblackburn1 (approved):

fblackburn1 (approved):

PR #216, opened 2021-09-02
WAZO-2193 confd add /sounds search/sort

ja0x:
Also added a new endpoint /sounds/{category}/files to allow sorting/searching on category's files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 21.12
+
+* The following query string parameters have been added to GET `/1.1/sounds`:
+
+  * `search`
+  * `order`
+  * `limit`
+  * `offset`
+  * `direction`
+
+
 ## 21.10
 
 * The name of call_permissions is now unique to a given tenant only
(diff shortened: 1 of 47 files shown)


fblackburn1 (changes requested):
I rebased your branch to fix copyright and fix the double quote usage

Please add copyright git-hook to your workflow:
https://github.com/wazo-platform/xivo-tools/tree/master/git-hooks

Please respect the service/wazo style. In all our code we use single quote (maybe some exception) and black is configured to respect this rule:
https://github.com/wazo-platform/wazo-confd/blob/master/tox.ini#L32
Always use tox to run black to avoid this kind of issue
 
When I read the ticket, I don't see any mention to create a new endpoint .. Why adding this, if not needed?
Moreover, before adding new endpoint/feature, we have an entire workflow to pass. Particularly, an validate the new openapi spec. I don't see any discussion about this here. Because `files` in the url is implicitly intended to return binary object, not metadata

Please split your branch to extract this new endpoint

fblackburn1 (approved):
Thanks, don't hesitate to ask if you have question about my commits

ja0x:
recheck

fblackburn1 (approved):

PR #217, opened 2021-09-02
Wazo 2506 meetings api

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/137
Depends-On: https://github.com/wazo-platform/xivo-bus/pull/42
Depends-On: https://github.com/wazo-platform/xivo-manage-db/pull/131


pc-m:
recheck

pc-m:
recheck

wazo_confd/plugins/meeting/resource.py
```diff
@@ -0,0 +1,258 @@
...
+        return body
+
+    def search_params(self):
+        params = ListSchema().load(request.args)
+        params['owner'] = self._find_user_uuid()
+        return params
+
+
+def random_string(length):
+    return ''.join(random.choice(string.ascii_lowercase) for _ in range(length))
```
sduthil, line 258:
```suggestion
    return ''.join(secrets.choice(string.ascii_lowercase) for _ in range(length))
```
This is exactly the purpose of the [`secrets` module](https://docs.python.org/3/library/secrets.html)

wazo_confd/plugins/meeting/resource.py
```diff
@@ -0,0 +1,258 @@
...
+            'endpoint_section_options': [
+                ['context', context],
+            ],
+            'identify_section_options': [],
+            'registration_section_options': [],
+            'registration_outbound_auth_section_options': [],
+            'outbound_auth_section_options': [],
+        }
+        endpoint_model = EndpointSIP(**endpoint_body)
+        endpoint = self._endpoint_sip_service.create(endpoint_model)
```
sduthil, line 106: I would prefer this to be in the service rather than the http module.
This is not a strong objection though.

wazo_confd/plugins/meeting/resource.py
```diff
@@ -0,0 +1,258 @@
...
+    def find_owners(self, form):
+        owner_uuids = form.pop('owner_uuids', None) or []
+        owners = []
+        tenant_uuid = form['tenant_uuid']
+        for uuid in owner_uuids:
+            owners.append(self._user_service.get(uuid, tenant_uuids=[tenant_uuid]))
+        form['owners'] = owners
+        return form
+
+    def _find_user_uuid(self):
```
sduthil, line 119: This is going to need some factorization: wazo-calld has the same code for the same purpose.

wazo_confd/plugins/meeting/resource.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @required_acl('confd.users.me.meetings.{uuid}.delete')
+    def delete(self, uuid):
+        kwargs = self._add_tenant_uuid()
+        kwargs['user_uuid'] = self._find_user_uuid()
+        model = self.get_model(uuid, **kwargs)
+        self.service.delete(model)
+        return '', 204
+
+    def _find_user_uuid(self):
```
sduthil, line 193: Same thing here.

wazo_confd/plugins/user/schema.py
```diff
@@ -17,7 +17,7 @@
 
 class UserSchema(BaseSchema):
     id = fields.Integer(dump_only=True)
-    uuid = fields.String(dump_only=True)
+    uuid = fields.String()
```
sduthil, line 20: This looks quite dangerous to me: could an API user change the UUID of a user after its creation?

sduthil (changes requested):
Thank you for all this :)

wazo_confd/plugins/meeting/resource.py
pc-m: I did not know about the secrets module. thanks

wazo_confd/plugins/meeting/resource.py
pc-m: I did it in the http module because we did that for the pjsip endpoint template logic (which I moved from the service to the http module)
I think @fblackburn1 could comment on that when he gets back. He will have a strong opinion ;)

sduthil (approved):

PR #218, opened 2021-09-07
WAZO-2488 Add exten to callfilter surrogate users

ja0x:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/138

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 21.12
 
+* The surrogates user list of `callfilters` resource now implements the `exten` field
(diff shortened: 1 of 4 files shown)


ja0x:
recheck

pc-m (changes requested):
Looks good, I've added some nitpicking comments about string formatting and I would like one more integration test where the bsfilter exten is disabled or not set.

You should also add the `exten` field to the open API spec and add a CHANGELOG entry for the change in CHANGELOG.md

pc-m (dismissed):

fblackburn1 (approved):

PR #220, opened 2021-09-23
refactor: use the user_uuid from the request

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-lib-python/pull/86

### DIFF
--- a/wazo_confd/plugins/meeting/resource.py
+++ b/wazo_confd/plugins/meeting/resource.py
@@ -8,7 +8,6 @@ import string
 from uuid import uuid4
 
 from flask import url_for, request
-from requests import HTTPError
 
 from xivo_dao.alchemy.meeting import Meeting
 from xivo_dao.alchemy.endpoint_sip import EndpointSIP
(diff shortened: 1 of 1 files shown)


pc-m:
recheck

pc-m:
recheck

pc-m:
recheck

wazo_confd/plugins/meeting/resource.py
```diff
@@ -190,20 +184,6 @@ def delete(self, uuid):
         self.service.delete(model)
         return '', 204
 
-    def _find_user_uuid(self):
-        token = request.headers.get('X-Auth-Token') or request.args.get('token')
```
sduthil, line 194: This is not the default behavior of `AuthVerifier`: it will only look at the header, not the query string.
`AuthVerifier` needs to be instanciated with `extract_token_id = extract_token_id_from_query_or_header`. This will introduce a slight change in behavior: the token from query string will take precedence over the header.

sduthil (changes requested):

wazo_confd/plugins/meeting/resource.py
sduthil: Actually we don't need the `token=` query string parameter so it's OK to remove it.

sduthil (approved):

PR #247, opened 2022-01-27
Use exten regex in relavant destinations

Romej:
## Summary of the changes

- use `EXTEN_REGEX` instead of `CONTEXT_REGEX` for `exten` field in extension destination
- use it for `exten` field in outcall destination too

Also, is it normal we don't offer the outcall destination in Portal-ui?


wazo_confd/helpers/destination.py
```diff
@@ -272,7 +272,7 @@ class NoneDestinationSchema(BaseDestinationSchema):
 class OutcallDestinationSchema(BaseDestinationSchema):
     outcall_id = fields.Integer(attribute='actionarg1', required=True)
     exten = fields.String(
-        validate=(Predicate('isdigit'), Length(max=255)),
+        validate=Regexp(EXTEN_REGEX),
```
fblackburn1, line 275: For now, I'll revert this commit:
1. We don't have real reason to change it
2. I don't know impact to allow `#*` in this field, and I'm quit lazy to spend time to investigate if we need it or not.
  * No objection to make this investigation if needed
3. This change behavior of the API by not allowing to set empty string
  * This is a good validation, but we should add migration script to avoid having empty string in database and make wrong logic based on this new validation

fblackburn1 (approved):
What is changed
* reverted latest commit (see above for explanation)
* updated copyright for `helpers/destinations.py`
* added non regression test

`mergeit` tag will be added after release and this will be merged automatically
Thanks for your contribution :pray:

wazo_confd/helpers/destination.py
Romej: All good, thank you for your feedback !

PR #248, opened 2022-02-01
WAZO-2500 Inject tenant_uuid in bus events

ja0x:


fblackburn1 (approved):
Works as expected :+1:

PR #249, opened 2022-02-08
Depends-On: https://github.com/wazo-platform/xivo-manage-db/pull/159

sduthil:
meetings: fail to create meeting if no SIP template

Why:

* It is misleading to return 201 if the meeting cannot be used by guests


pc-m (approved):

pc-m:
recheck

pc-m (approved):

PR #253, opened 2022-03-02
Change voicemail "number" type to integer

Romej:
## Summary of the changes
- voicemail `number` type changed to `integer` in schema and api spec 

## Motivation
- found it confusing that we expect a string for this property, only to then apply a number regex on it


fblackburn1 (commented):
Be careful, in telephony number/exten != integer
ex:
`001` is a valid number
but if you cast it to integer, you will have `1`

PR #256, opened 2022-03-15
[WAZO-2674] event_handlers: delete devices on tenant deletion

bloom1:


wazo_confd/plugins/event_handlers/plugin.py
```diff
@@ -25,10 +32,25 @@ def _auth_tenant_added(self, event):
...
             self.service.copy_slug(tenant, slug)
 
+    def _auth_tenant_deleted(self, event):
+        tenant_uuid = event['uuid']
+        logger.debug('Deleting devices for tenant "%s"', tenant_uuid)
+        devices = self.provd.devices.list(tenant_uuid=tenant_uuid, recurse=True)[
+            'devices'
+        ]
+        for device in devices:
+            self.provd.devices.delete(device['id'])
```
sduthil, line 42: This code only deleted the device, but the config should be deleted too, and an event should be sent, as done in the [devices service](https://github.com/wazo-platform/wazo-confd/blob/master/wazo_confd/plugins/device/service.py#L45-L47).

sduthil (changes requested):

sduthil (approved):

PR #257, opened 2022-03-25
pin jinja2 and marksupsafe to buster version

fblackburn1:
why: latest jinja2 version (3.1.0) is incompatible with flask 1.0.2
(from buster)

and newer markupsafe versions (> 2) are incompatible with buster jinja2


fblackburn1:
recheck

fblackburn1:
recheck

PR #259, opened 2022-04-12
Pull request for WAZO-2750-purge-meeting-authorization

sduthil:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/172

## meeting purge: fix timezone

Why:

* datetime.utcnow() returns UTC time WITHOUT a timezone
* when comparing the database date with a timestamp without a
timezone, PGSQL does the comparison in the local time zone
* when comparing with both timezones, PGSQL applies the corresponding
offset before comparing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 22.06
+
+* Meeting authorizations are now only valid for 24 hours. They are automatically
+  deleted after 24 hours.
+
 ## 22.05
 
 * The following resources have been added:
--- a/debian/cron.d
+++ b/debian/cron.d
@@ -2,5 +2,5 @@
 # cron jobs for wazo-confd
 #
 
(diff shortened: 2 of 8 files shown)


fblackburn1 (commented):
What do you think about merging this script with purge-meetings and only add a flag `--authorizations-only` ? It will remove duplicate overhead

fblackburn1 (changes requested):
Tests are missing
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/172

sduthil:
I've added the test and merge the purge script with `wazo-confd-purge-meetings`. Thanks for the review!

sduthil:
recheck

fblackburn1 (approved):
IMHO a really not perfect but more readable way to write many context managers with black constraint:
https://github.com/wazo-platform/wazo-confd/blob/master/integration_tests/suite/base/test_extensions.py#L543-L550

PR #290, opened 2022-09-13
Wazo 2864 add lines to user

pc-m:


wazo_confd/plugins/user/api_v2_0.yml
```diff
@@ -0,0 +1,313 @@
...
+      agent:
+        readOnly: true
+        $ref: '#/definitions/AgentRelationBase'
+  UserRequest:
+    title: Unified User
+    properties:
+        user:
+          type: object
+          description: User Section
+          $ref: '#/definitions/User'
```
sduthil, line 252: IMHO this is incompatible with the `/2.0/users` pattern: if the resource is a user, then the fields of this resource should be first-level, e.g. `{firstname: ..., lines: {...}}`

For me `/2.0/...` is the future of the confd API, and all resources will eventually move to `/2.0/...`, so it needs to be ready for more resources.

Having `users: {...}` would be correct for a URL like `/unified/users` that exists *besides* `/users`, but not for `/2.0/users`: either we'll have to keep the same pattern for other resources (e.g. `/2.0/queues` will also have a `queues: {...}` field), or we will break homogeneity between all resources.

And if we remove the `user` key, then do we really need `/2.0`, i.e. just add new keys in `POST /1.1/users`?

wazo_confd/plugins/line/resource.py
```diff
@@ -51,3 +54,42 @@ def put(self, id):
     @required_acl('confd.lines.{id}.delete')
     def delete(self, id):
         return super().delete(id)
+
+
+# class LineListV2(ListResource):
```
sduthil, line 59: Why is this commented out?

wazo_confd/plugins/user/resource.py
```diff
@@ -61,3 +74,72 @@ def put(self, id):
...
+        for line_body in body.get('lines') or []:
+            extensions = line_body.pop('extensions', None) or []
+            endpoint_sip = line_body.pop('endpoint_sip')
+            endpoint_sip, _, _ = self._endpoint_sip_list_resource._post(endpoint_sip)
+            line, _, _ = self._line_list_resource._post(line_body)
+            self._line_endpoint_sip_association_resource.put(line['id'], endpoint_sip['uuid'])
+            line['endpoint_sip'] = endpoint_sip
+            line['extensions'] = []
+            for extension_body in extensions:
+                extension, _, _ = self._extension_line_list_resource._post(line['id'], extension_body)
```
sduthil: What happens if the extension creation fails? Can you confirm that nothing is committed in database?

wazo_confd/plugins/user/schema.py
```diff
@@ -229,3 +240,238 @@ def on_bind_field(self, field_name, field_obj):
...
+    )
+
+    fallbacks = Nested('UserFallbackSchema', dump_only=True)
+    groups = Nested(
+        'GroupSchema', only=['uuid', 'id', 'name', 'links'], many=True, dump_only=True
+    )
+    incalls = Nested(
+        'IncallSchema', only=['id', 'extensions', 'links'], many=True, dump_only=True
+    )
+    lines = Nested(
```
sduthil, line 306: Why do we keep `.user.lines` besides the `.lines` key?

wazo_confd/plugins/user/schema.py
```diff
@@ -229,3 +240,238 @@ def on_bind_field(self, field_name, field_obj):
...
+    @classmethod
+    def _flatten(cls, iterable_of_iterable):
+        for item in iterable_of_iterable:
+            try:
+                itercheck = iter(item)
+                yield from cls._flatten(itercheck)
+            except TypeError:
+                yield item
+
+    # DEPRECATED 20.01
```
sduthil, line 364: Do we really need to keep the deprecated code in 2.0?

wazo_confd/plugins/user/schema.py
```diff
@@ -229,3 +240,238 @@ def on_bind_field(self, field_name, field_obj):
...
+                break
+            else:
+                raise ValidationError('A line or it\'s extension need a "context"')
+
+        for line in lines:
+            if not line.get('endpoint_sip'):
+                raise ValidationError('A line needs a SIP endpoint')
+
+    @classmethod
+    def _flatten(cls, iterable_of_iterable):
```
sduthil, line 471: I don't understand the purpose of this method. Is it called somewhere?

PR #291, opened 2022-09-20
Give confd more time/tries to restart

nissou91:
**Description:**
* confd takes much more time to start after it has been shut down and restarted. We need to increase the timeout and give it more chances/tries.

### DIFF
--- a/integration_tests/suite/base/test_status.py
+++ b/integration_tests/suite/base/test_status.py
@@ -95,17 +95,20 @@ def test_confd_status_fails_when_wazo_auth_is_down():
     confd._reset()
     confd_csv._reset()
 
-    response = until.return_(confd.status.get, timeout=5)
-    assert_that(response.item, has_entries(**expected_status_fail_entries))
+    def _wait_until_no_exception_is_raised():
+        response = until.return_(confd.status.get, timeout=10)
+        assert_that(response.item, has_entries(**expected_status_fail_entries))
+
+    until.assert_(_wait_until_no_exception_is_raised, tries=10)
 
     BaseIntegrationTest.start_service('auth')
     auth._reset()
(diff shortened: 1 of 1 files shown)


nissou91:
recheck

integration_tests/suite/base/test_status.py
```diff
@@ -95,17 +95,20 @@ def test_confd_status_fails_when_wazo_auth_is_down():
     confd._reset()
     confd_csv._reset()
 
-    response = until.return_(confd.status.get, timeout=5)
-    assert_that(response.item, has_entries(**expected_status_fail_entries))
+    def _wait_until_no_exception_is_raised():
+        response = until.return_(confd.status.get, timeout=10)
+        assert_that(response.item, has_entries(**expected_status_fail_entries))
+
+    until.assert_(_wait_until_no_exception_is_raised, tries=10)
```
fblackburn1, line 102: Why changing this part of code?
Wrapping code inside a `until.assert_` won't change anything with the `ConnectionError` from timeout
Did you check the `assert_()` implementation? It won't catch other thing than `AssertionError`

fblackburn1 (changes requested):

integration_tests/suite/base/test_status.py
nissou91: Also the reason I increased the wait time for confd is because when the tests are run separately on the stack, they succeed at each time, however when we run these tests on zuul, they succeed sometimes and sometimes they fail. The only plausible reason that I see for that is the singleton shared object needs more time to be reset; I do not think that it is caused anywhere else in the code...

nissou91:
recheck

sduthil (approved):

PR #298, opened 2022-10-03
Restore the /users with lines, extensions and SIP endpoint

pc-m:
This PR restores the new user API branch that was reverted because it created a wazo-auth user given body that used to work in previous version and did NOT create one.

This new version will do the following.

1. If a line is created and its body includes an endpoint (SIP, SCCP or custom) the endpoint will be created and associated to the line at the same time.
2. If the line has an extension in its body when it gets created, then the extension will be created and associated while creating the line.
3. If the user has a line in its body when it gets created, then the line will be created and associated while creating the user.
4. If an `auth` section is in the user's body when creating the user, then the wazo-auth user for that user will be created at the same time as the wazo-confd user.

When an error occurs nothing should happen. I did not write a test for a stopped wazo-auth since that would have returned a 503 from the token checking mechanism and not from the implementation of this resource anyway.

### DIFF
--- a/wazo_confd/plugins/line/resource.py
+++ b/wazo_confd/plugins/line/resource.py
@@ -1,20 +1,81 @@
-# Copyright 2015-2020 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2015-2022 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-from flask import url_for, request
+from flask import url_for
 
 from xivo_dao.alchemy.linefeatures import LineFeatures as Line
 
 from wazo_confd.auth import required_acl
 from wazo_confd.helpers.restful import ListResource, ItemResource
-from wazo_confd.plugins.line.schema import LineSchema, LineSchemaNullable
+from wazo_confd.plugins.line.schema import (
+    LineSchemaNullable,
+    LineListSchema,
+    LinePutSchema,
+)
+
+from wazo_confd.plugins.endpoint_custom.resource import CustomList
+from wazo_confd.plugins.endpoint_sccp.resource import SccpList
+from wazo_confd.plugins.endpoint_sip.resource import SipList
+from wazo_confd.plugins.line_endpoint.resource import (
+    LineEndpointAssociationCustom,
+    LineEndpointAssociationSccp,
+    LineEndpointAssociationSip,
+)
+from wazo_confd.plugins.line_extension.resource import LineExtensionList
 
 
 class LineList(ListResource):
 
     model = Line
     schema = LineSchemaNullable
+    _list_schema = LineListSchema
     has_tenant_uuid = True
+
+    def __init__(
+        self,
+        line_service,
+        endpoint_custom_service,
+        endpoint_sip_service,
+        extension_line_service,
+        extension_service,
+        line_endpoint_custom_association_service,
+        line_endpoint_sip_association_service,
+        line_endpoint_sccp_association_service,
+        endpoint_sccp_service,
+        endpoint_custom_dao,
+        endpoint_sccp_dao,
+        line_dao,
+        sip_dao,
+        transport_dao,
+    ):
+        super().__init__(line_service)
+        self._endpoint_sip_list_resource = SipList(
+            endpoint_sip_service, sip_dao, transport_dao
+        )
+        self._endpoint_sccp_list_resource = SccpList(
+            endpoint_sccp_service,
+        )
+        self._endpoint_custom_list_resource = CustomList(
+            endpoint_custom_service,
+        )
+        self._line_endpoint_custom_association_resource = LineEndpointAssociationCustom(
+            line_endpoint_custom_association_service,
+            line_dao,
+            endpoint_custom_dao,
+        )
+        self._line_endpoint_sip_association_resource = LineEndpointAssociationSip(
+            line_endpoint_sip_association_service, line_dao, sip_dao
+        )
+        self._line_endpoint_sccp_association_resource = LineEndpointAssociationSccp(
+            line_endpoint_sccp_association_service,
+            line_dao,
+            endpoint_sccp_dao,
+        )
+        self._extension_line_list_resource = LineExtensionList(
+            extension_line_service,
+            extension_service,
+            line_dao,
+        )
 
     def build_headers(self, line):
         return {'Location': url_for('lines', id=line.id, _external=True)}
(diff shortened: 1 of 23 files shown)


wazo_confd/plugins/user/resource.py
```diff
@@ -92,11 +93,9 @@ def _post(self, body):
             self._user_line_item_resource.put(user_dict['uuid'], line['id'])
             user_dict['lines'].append(line)
 
-        # FIX: create(...) takes a user dict containing an 'email_address' key, not an 'email' key
-        if user_dict.get('username'):
-            fixed_user_dict = user_dict.copy()
-            fixed_user_dict['email_address'] = fixed_user_dict['email']
-            self._wazo_user_service.create(fixed_user_dict)
+        if auth:
```
fblackburn1, line 96: I didn't read the PR (or previous) and context, but according old spec and user import, the wazo-auth user should always be created, it is a part of the contract
With this logic, it will be a headache to build UI to fix confd user with auth user

pc-m: If wazo-auth is unreachable a 503 is returned

pc-m: The auth user is only created when the `auth` section is specified in the body. If someone uses a the body from a previous version the auth user is not created so the behavior stays the same as before

wazo_confd/plugins/line/schema.py
```diff
@@ -29,9 +31,55 @@ class LineSchema(BaseSchema):
...
+
+class LineSchemaNullable(LineSchema):
+    def on_bind_field(self, field_name, field_obj):
+        super().on_bind_field(field_name, field_obj)
+        nullable_fields = ['provisioning_code', 'position', 'registrar']
+        if field_name in nullable_fields:
+            field_obj.allow_none = True
+
+
+class LineListSchema(LineSchemaNullable):
```
fblackburn1, line 75: I may miss something, but I don't understand why you need to inherit from Nullable schema here?
LineListSchema is (or should) only be used with `dump` but allowing none is a restriction when we do `load`

wazo_confd/plugins/user/api.yml
```diff
@@ -288,7 +288,30 @@ parameters:
...
+    - $ref: '#/definitions/UserRelationAgent'
+    - $ref: '#/definitions/UserRelationFallbacks'
+    - $ref: '#/definitions/UserRelationForwards'
+    - $ref: '#/definitions/UserRelationGroups'
+    - $ref: '#/definitions/UserRelationIncalls'
+    - $ref: '#/definitions/UserRelationLinesCreate'
+    - $ref: '#/definitions/UserRelationServices'
+    - $ref: '#/definitions/UserRelationSwitchboards'
+    - $ref: '#/definitions/UserRelationQueues'
+    - $ref: '#/definitions/UserRelationCallPickupTargets'
```
fblackburn1, line 305: Not everything is implemented in this PR

wazo_confd/plugins/user/resource.py
```diff
@@ -1,34 +1,105 @@
...
 from .schema import (
     UserDirectorySchema,
-    UserSchema,
+    UserListSchema,
     UserSchemaNullable,
     UserSummarySchema,
+    UserPutSchema,
 )
 
+logger = logging.getLogger(__name__)
```
fblackburn1, line 23: unused logger

fblackburn1 (changes requested):
It's a pure code review
I will make functional tests and corner cases in another review

wazo_confd/plugins/line/resource.py
```diff
@@ -25,16 +86,64 @@ def get(self):
...
+                model.id,
+                endpoint_custom['id'],
+            )
+            payload['endpoint_custom'] = endpoint_custom
+
+        for extension_body in extensions:
+            extension, _, _ = self._extension_line_list_resource._post(
+                model.id, extension_body
+            )
+            payload['extensions'].append(extension)
```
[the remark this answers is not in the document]
pc-m: I agree that `service.py` should be the place to do most of this logic.

The current implementation in restful does many things
1. Extracts the payload from the request
2. Deserializes the payload
3. Add the `tenant_uuid` to the payload when appropriate
4. Instantiate the sqlalchemy model from the payload
5. Calls `service.create`
6. Serializes the response
7. Return the response, status code and headers

I could extract steps 4 and move it into the service. That would allow the input to be inspected or modified before instantiating the sqlalchemy model.

I "think" (I would have to try with a resource first) that that would allow me to move all logic from the resources to the services. Also it would avoid the necessity of splitting the `post` in 2 functions in `restful.py`

I would also like to introduce a new class `ServiceHandle` that would reference all services that could be added to the `dependencies` and allow multiple plugins to use a service without having to instantiate the service and import all of its dependencies.

That would be used like that
```diff
diff --git a/wazo_confd/plugins/extension/plugin.py b/wazo_confd/plugins/extension/plugin.py
index 77d5125a..91fce513 100644
--- a/wazo_confd/plugins/extension/plugin.py
+++ b/wazo_confd/plugins/extension/plugin.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2019 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2016-2022 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 from wazo_provd_client import Client as ProvdClient
@@ -12,11 +12,13 @@ class Plugin:
         api = dependencies['api']
         config = dependencies['config']
         token_changed_subscribe = dependencies['token_changed_subscribe']
+        service_handle = dependencies['service_handle']
 
         provd_client = ProvdClient(**config['provd'])
         token_changed_subscribe(provd_client.set_token)
 
         service = build_service(provd_client)
+        service_handle.register('extension', service)
 
         api.add_resource(
             ExtensionItem,
```

```diff
diff --git a/wazo_confd/plugins/line/resource.py b/wazo_confd/plugins/line/resource.py
index 5230bf39..190b7a75 100644
--- a/wazo_confd/plugins/line/resource.py
+++ b/wazo_confd/plugins/line/resource.py
@@ -19,6 +19,10 @@ class LineList(ListResource):
     def build_headers(self, line):
         return {'Location': url_for('lines', id=line.id, _external=True)}
 
+    def __init__(self, line_service, service_handle):
+        super().__init__(line_service)
+        self._extension_service = service_handle.get('extension')
+
     @required_acl('confd.lines.read')
     def get(self):
         return super().get()
```
```diff
diff --git a/wazo_confd/controller.py b/wazo_confd/controller.py
index 23ff58df..de65893e 100644
--- a/wazo_confd/controller.py
+++ b/wazo_confd/controller.py
@@ -24,6 +24,18 @@ from .service_discovery import self_check
 logger = logging.getLogger(__name__)
 
 
+class ServiceHandle:
+
+    def __init__(self):
+        self._services = {}
+
+    def register(self, name, service):
+        self._services[name] = service
+
+    def service(self, name):
+        return self._services[name]
+
+
 class Controller:
     def __init__(self, config):
         self.config = config
@@ -51,6 +63,7 @@ class Controller:
         self.token_renewer.subscribe_to_token_change(
             self.token_status.token_change_callback
         )
+        service_handle = ServiceHandle()
         self.status_aggregator.add_provider(auth.provide_status)
         self.status_aggregator.add_provider(self.token_status.provide_status)
         self.status_aggregator.add_provider(self._bus_consumer.provide_status)
@@ -66,6 +79,7 @@ class Controller:
                 'bus_publisher': self._bus_publisher,
                 'auth_client': auth_client,
                 'pjsip_doc': pjsip_doc,
+                'service_handle': service_handle,
                 'status_aggregator': self.status_aggregator,
             },
         )
```

PR #299, opened 2022-10-06
Wazo 2887 user fallbacks

pc-m:


xxbillouxx:
recheck

wazo_confd/plugins/user/resource.py
```diff
@@ -1,34 +1,161 @@
...
+
     def build_headers(self, user):
         return {'Location': url_for('users', id=user.id, _external=True)}
 
     @required_acl('confd.users.create')
     def post(self):
         return super().post()
 
+    def _post(self, body):
+        # forwards payload will be modified by the PUT body
```
xxbillouxx, line 112: why is it before `body = self.schema().load(body)` ?

xxbillouxx (commented):
gj

wazo_confd/plugins/user/resource.py
pc-m: The schema forwards schema modifies the data. If I get the forwards after the `load` the forwards resource will try to modify the schema again. This returns a `400`

PR #302, opened 2022-10-25
Fix dissociation on line extension delete

pc-m:
This is a followup to https://github.com/wazo-platform/xivo-manage-db/pull/170/files
Since the CASCADE does not execute the line_extension fixes this dissociate the line/extension explicitly when deleting a line or extension that are associated.

I have not been able to spot a bug to would result from not executing the fixes


fblackburn1:
No ticket and no reviewer assigned. Is this branch still valid?

PR #322, opened 2022-11-14
Add middlewares for /users

xxbillouxx:


wazo_confd/plugins/user/middleware.py
```diff
@@ -1,22 +1,103 @@
...
+            Session.expire(model)
+
+        if groups:
+            self._middleware_handle.get('user_group_association').associate_all_groups(
+                {'groups': groups}, user_dict['uuid']
+            )
+            Session.expire(model)
+        user_dict['groups'] = groups
+
+        if switchboards:
```
pc-m, line 85: You don't need the `if` here since there's only a `for`. Otherwise you can remove the `or []` when popping switchboards

wazo_confd/plugins/user/middleware.py
```diff
@@ -1,22 +1,103 @@
...
+        if lines:
+            for line_body in lines:
+                line = self._middleware_handle.get('line').create(
+                    line_body, tenant_uuid, tenant_uuids
+                )
+                self._middleware_handle.get('user_line_association').associate(
+                    user_dict['id'], line['id'], tenant_uuids
+                )
+                user_dict['lines'].append(line)
+            Session.expire(model)
```
pc-m, line 50: Could we use `Session.expire(model, ['lines'])` here instead of expiring everything?

wazo_confd/plugins/user/schema.py
```diff
@@ -229,3 +243,37 @@ def on_bind_field(self, field_name, field_obj):
...
+class UserListItemSchema(UserSchemaNullable):
+    auth = Nested('WazoAuthUserSchema')
+    lines = Nested('LineSchema', many=True)
+    incalls = Nested('UserIncallSchema', many=True)
+    groups = Nested('UserGroupSchema', many=True)
+    switchboards = Nested('UserSwitchboardSchema', many=True)
+
+
+class UserIncallSchema(BaseSchema):
+    id = fields.Integer(dump_only=True)
```
pc-m, line 257: Most fields from the `incall` schema should be available here.
`preprocess_subroutine`, `greeting_sound`, `caller_id_mode`, `caller_id_name`, `description`

integration_tests/suite/base/test_user_import.py
```diff
@@ -107,40 +107,42 @@ def test_given_csv_has_all_fields_for_a_user_then_user_imported():
...
-            userfield="userfield",
-            call_permission_password='1234',
-            enabled=True,
-            uuid=user_uuid,
-            username=None,
-            password=None,
-            subscription_type=2,
-        ),
-    )
+    try:
```
pc-m, line 110: You should move the `try` after assigning `user_uuid` otherwise the finally will fail adding noise to the test.

pc-m (changes requested):

pc-m:
also api.yml and the changelog need to be modified

xxbillouxx:
recheck

xxbillouxx:
recheck

xxbillouxx:
recheck

pc-m (approved):

xxbillouxx:
recheck

xxbillouxx:
recheck

PR #329, opened 2022-11-24
Add middleware for user-agent

xxbillouxx:


xxbillouxx:
recheck

xxbillouxx:
recheck

xxbillouxx:
recheck

xxbillouxx:
recheck

integration_tests/suite/base/test_users.py
```diff
@@ -915,6 +915,7 @@ def test_post_full_user_no_error(
         group,
         switchboard,
     ) = generate_user_resources_bodies(group, switchboard)
+    agent = {}
```
pc-m, line 918: oh shit, I did not anticipate a body that would be THAT small. When you know about the implementation it makes sense but as a user I would expect a boolean.

What would you think of allowing `"agent": true` that would be equivalent to `"agent": {}`
The following values would be possible.
true
{"firstname": "Nabil", "number": "666"}
null
and false I guess

@fblackburn1 Do you have an idea about this one?

pc-m (changes requested):
api.yml and changelog missing

integration_tests/suite/base/test_users.py
xxbillouxx: I think that a field having 2 differents types (dict and boolean) is ambiguous.
What's opinion @fblackburn1 ?

fblackburn1: Ooh well .. I agree with @xxbillouxx 
But understand that `agent: {}` may be not obvious.
Maybe a key helper `with_agent: true`? (only for the POST body, not for the returned body) - (and not required to be in this PR/scope)

Anyways, not a strong opinion, do what you want :smile:

pc-m: Ok then we are going to do whatever @xxbillouxx wants here :)
Also I'm going to take a screenshot of @fblackburn1 not having a strong opinion :rofl:

pc-m (approved):

PR #336, opened 2022-12-09
devices: do not raise exception on missing configuration key

bloom1:
Why: this would make it impossible to list the devices


integration_tests/suite/base/test_devices.py
```diff
@@ -159,6 +159,24 @@ def test_list_unallocated(main, sub):
...
 
 
+@fixtures.device()
+@fixtures.device()
+def test_list_device_no_config(dev1, dev2):
+    real_device = provd.devices.get(dev1['id'])
+    real_device['config'] = None
+    provd.devices.update(real_device)
+
+    new_dev2 = confd.devices(dev2['id']).get().item
```
pc-m, line 169: Hi @bloom1
My morning brain refuses to understand what `new_dev2` is used for in the test. Is it different from `dev2`? Does GETing it tests a side effect of fixing the bug?

pc-m (approved):

PR #350, opened 2023-01-18
Delete voicemail is not used by users

xxbillouxx:


wazo_confd/plugins/user_voicemail/middleware.py
```diff
@@ -24,3 +25,15 @@ def create_voicemail(self, user_id, body, tenant_uuids):
...
+    def delete_voicemail(self, user_id, tenant_uuids):
+        user = user_dao.get_by_id_uuid(user_id, tenant_uuids=tenant_uuids)
+        voicemail_middleware = self._middleware_handle.get('voicemail')
+        voicemail = voicemail_middleware.get(user.voicemailid, tenant_uuids)
+
+        self.dissociate(user_id, tenant_uuids)
+        Session.expire(user, ['voicemail'])
+
+        # if there is no other user using the voicemail
+        if len(voicemail['users']) == 1:
```
pc-m, line 38: I don't like the `len == 1` to check if it's not used anymore. I had to carefully look at the context to check that is was correct.
How about this
```diff
diff --git a/wazo_confd/plugins/user_voicemail/middleware.py b/wazo_confd/plugins/user_voicemail/middleware.py
index ae367e68..99f17a4e 100644
--- a/wazo_confd/plugins/user_voicemail/middleware.py
+++ b/wazo_confd/plugins/user_voicemail/middleware.py
@@ -29,11 +29,11 @@ class UserVoicemailMiddleware:
     def delete_voicemail(self, user_id, tenant_uuids):
         user = user_dao.get_by_id_uuid(user_id, tenant_uuids=tenant_uuids)
         voicemail_middleware = self._middleware_handle.get('voicemail')
-        voicemail = voicemail_middleware.get(user.voicemailid, tenant_uuids)
+        voicemail_id = user.voicemailid
 
         self.dissociate(user_id, tenant_uuids)
         Session.expire(user, ['voicemail'])
 
-        # if there is no other user using the voicemail
-        if len(voicemail['users']) == 1:
+        voicemail = voicemail_middleware.get(voicemail_id, tenant_uuids)
+        if not voicemail['users']:
             voicemail_middleware.delete(voicemail['id'], tenant_uuids)
```

pc-m (changes requested):
CHANGELOG is missing but I think we could add it at the end of all the DELETE to avoid rebasing for each PR

pc-m (approved):

PR #358, opened 2023-01-24
Enable update on user agent

xxbillouxx:


pc-m (dismissed):

wazo_confd/plugins/user/middleware.py
```diff
@@ -396,3 +397,17 @@ def update(self, user_id, body, tenant_uuid, tenant_uuids, recursive=False):
                     tenant_uuid,
                     tenant_uuids,
                 )
+
+            if agent:
+                agent_id = user.agentid
+                self._middleware_handle.get('user_agent_association').dissociate(
+                    user_id, tenant_uuids
+                )
+                self._middleware_handle.get('agent').delete(agent_id, tenant_uuids)
```
pc-m, line 406: Doing a delete/create will generate a new agent id. This would probably break some stuff. The first thing that comes to mind is the agent statistics which are based on the agent's ID. Doing a delete/create would either leave some orphaned agent statistics or delete them, both are bad.

pc-m (changes requested):

wazo_confd/plugins/user/middleware.py
```diff
@@ -400,14 +400,19 @@ def update(self, user_id, body, tenant_uuid, tenant_uuids, recursive=False):
...
-                self._middleware_handle.get('user_agent_association').dissociate(
-                    user_id, tenant_uuids
-                )
-                self._middleware_handle.get('agent').delete(agent_id, tenant_uuids)
+                if agent_id:
+                    self._middleware_handle.get('agent').update(
+                        agent_id, agent, tenant_uuids
+                    )
+                else:
+                    self._middleware_handle.get('user_agent_association').dissociate(
```
pc-m, line 408: If user has no agentid what are we dissociating from here?

wazo_confd/plugins/user/middleware.py
```diff
@@ -400,14 +400,19 @@ def update(self, user_id, body, tenant_uuid, tenant_uuids, recursive=False):
...
+                        user_id, tenant_uuids
+                    )
+                    self._middleware_handle.get('agent').delete(agent_id, tenant_uuids)
```
pc-m, line 411: And what is being deleted here?

pc-m (approved):

pc-m (approved):

PR #361, opened 2023-01-26
[WP-624] event_handlers: add new default endpoint option

bloom1:
`notify_early_inuse_ringing` is available to mimick the chan_sip
behaviour for BLF function keys to pickup a call of a user that is
already busy.


pc-m (approved):

bloom1:
recheck

PR #368, opened 2023-02-09
Enable update on user groups

xxbillouxx:


sduthil:
recheck

wazo_confd/plugins/user/middleware.py
```diff
@@ -352,3 +356,8 @@ def update(self, user_id, body, tenant_uuids, recursive=False):
                 self._middleware_handle.get('user_forward_association').associate(
                     user_id, forwards
                 )
+
+            if groups:
```
pc-m, line 360: If the user is associated to a group and I want to remove all group associations with a PUT I would do
```javascript
{
"groups": [],
}
```
But that would not remove any of my current group associations

pc-m (changes requested):

pc-m (approved):

PR #369, opened 2023-02-10
Create user with voicemail using user name

xxbillouxx:


wazo_confd/plugins/user/middleware.py
```diff
@@ -54,6 +54,12 @@ def create(self, body, tenant_uuid, tenant_uuids):
                 )
                 voicemail = voicemail_middleware.get(voicemail_id, tenant_uuids)
             else:
+                if (
+                    not voicemail.get('name', None)
```
pc-m, line 58: Since the user's first name is mandatory and not the last name if should be something like
```python
if not voicemail.get('name'):
    voicemail['name'] = f'{model.firstname} {model.lastname}' if model.lastname else f'{model.firstname}'
```

pc-m (changes requested):

PR #374, opened 2023-02-21
Wp 855 user update lines

xxbillouxx:


wazo_confd/plugins/extension/middleware.py
```diff
@@ -28,3 +28,23 @@ def get(self, extension_id, tenant_uuids):
     def get_by(self, tenant_uuids, **criteria):
         model = self._service.get_by(tenant_uuids=tenant_uuids, **criteria)
         return self._schema.dump(model)
+
+    def parse_and_update(self, model, body, **kwargs):
```
pc-m, line 32: parse_and_update and find_updated_fields will end up in every middleware where we do an update. Move these functions to a base class or to a helper function

wazo_confd/plugins/line/middleware.py
```diff
@@ -119,3 +122,99 @@ def delete(self, line_id, tenant_uuids, recursive=False):
...
+        if extension_bodies:
+            line_extension_middleware = self._middleware_handle.get('line_extension')
+            existing_extensions = {ext.id: ext for ext in model.extensions}
+
+            extensions_body_to_be_updated = {}
+            extensions_body_to_be_created = []
+            extensions_body_to_be_deleted = []
+            for extension_body in extension_bodies:
+                try:
+                    extensions_body_to_be_updated[extension_body['id']] = extension_body
```
pc-m, line 190: Is the following case handled somewhere?
```javascript
{'exten': '1004', 'context': 'default', 'id': null}
```

pc-m (changes requested):

integration_tests/suite/base/test_users.py
```diff
@@ -1776,3 +1792,305 @@ def test_post_incalls_existing_extension_wrong_context_type_error(outcall, exten
...
+        _,
+        line,
+        _,
+        _,
+        _,
+        _,
+        _,
+        _,
+        _,
+    ) = generate_user_resources_bodies(
```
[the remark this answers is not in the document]
xxbillouxx: I created a dedicated class for that

wazo_confd/plugins/extension/middleware.py
xxbillouxx: I created a base class for the middlewares

wazo_confd/plugins/line/middleware.py
xxbillouxx: I updated the schema and added a test

pc-m (approved):

PR #377, opened 2023-02-24
Enable update on user voicemail

xxbillouxx:


wazo_confd/plugins/user_group/validator.py
```diff
@@ -8,7 +8,9 @@
 
 class UserGroupAssociationValidator(ValidatorAssociation):
     def validate(self, user, groups):
-        self.validate_user_has_endpoint(user)
+        # if there is no group to associate to the user, lines are not required
+        if len(groups):
```
pc-m, line 12: Just nitpicking, you can use `if groups:` instead of counting

pc-m (changes requested):

pc-m (approved):

PR #381, opened 2023-03-14
Update OpenAPI specs for user creation

xxbillouxx:


DrPyser:
I'm confused about how the openapi docs are/should be organized between plugins­.

DrPyser (approved):
Verified this doc against schemas, seems good.

PR #383, opened 2023-03-29
Allow many lines wIth the same extension

xxbillouxx:


integration_tests/suite/base/test_users.py
```diff
@@ -2417,3 +2417,59 @@ def test_update_incall_new_incall():
         ),
     )
     url.delete(recursive=True)
+
+
+@fixtures.device()
+@fixtures.device()
+def test_post_lines_same_unallocated_device_no_error(device, device2):
```
fblackburn1, line 2424: Why testing device here? The use case is not "on the same device" but "on the same extension"

integration_tests/suite/base/test_users.py
```diff
@@ -2417,3 +2417,59 @@ def test_update_incall_new_incall():
...
+    }
+    line2['device_id'] = device2['id']
+
+    # user creation with 2 lines
+    response = confd.users.post(
+        {
+            'lines': [user_resources.line, line2],
+            **user_resources.user,
+        }
+    )
```
fblackburn1, line 2441: Writing the code differently avoid to write comment. Example, if you want to focus on your lines, you can separate this on a specific variable
```suggestion
    lines = [user_resources.line, line2]
    response = confd.users.post(lines=lines, **user_resources.user)
```
**Note: Not 100% if what inside of `user_resources.user`  is compatible with the kwargs, if not, you can re-embedded into a dictionary as before

integration_tests/suite/base/test_users.py
```diff
@@ -2417,3 +2417,59 @@ def test_update_incall_new_incall():
...
+
+    # retrieve the lines (created before) and check if the device is associated to the line
+    assert_that(
+        confd.lines(payload['lines'][0]['id']).get().item,
+        has_entries(device_id=device['id']),
+    )
+    assert_that(
+        confd.lines(payload['lines'][1]['id']).get().item,
+        has_entries(device_id=device2['id']),
+    )
```
fblackburn1, line 2471: Prefer code styling instead of comment. ex:
```suggestion
    existing_line_1 = confd.lines(payload['lines'][0]['id']).get().item
    added_device_id_1 = device['id']
    assert_that(existing_line_1, has_entries(device_id=added_device_id_1))
    
    existing_line_2 = confd.lines(payload['lines'][1]['id']).get().item
    added_device_id_2 = device2['id']
    assert_that(existing_line_2, has_entries(device_id=added_device_id_2))
```
**Note: feel free to change name - not tested
**Note: Refering to previous comment, I'm not sure we should check device_id instead of extension

fblackburn1 (changes requested):
Code work as expected :+1: 
Only few questions/refactors on test side

integration_tests/suite/base/test_users.py
xxbillouxx: function renamed

integration_tests/suite/base/test_users.py
xxbillouxx: comment removed 
and dict param removed for post(...)

integration_tests/suite/base/test_users.py
xxbillouxx: assert on extension instead of device and comment removed

fblackburn1 (approved):

PR #385, opened 2023-04-17
Block extension deletion if used

xxbillouxx:


wazo_confd/plugins/user/middleware.py
```diff
@@ -439,7 +445,11 @@ def create_or_update_extension(extension_id, extension_body, tenant_uuids):
         # all the extensions (that have not been processed previously)
         # are not used anymore so can be removed
         for e in existing_extensions_ids:
-            self._middleware_handle.get('extension').delete(e, tenant_uuids)
+            try:
+                self._middleware_handle.get('extension').delete(e, tenant_uuids)
+            except ResourceError as e:
+                if not str(e).startswith('Resource Error - Extension is associated'):
+                    raise e
```
DrPyser, line 452: (see comment about exception subclass)

wazo_confd/plugins/line_extension/middleware.py
```diff
@@ -37,4 +37,8 @@ def delete_extension(self, line_id, extension_id, tenant_uuid, tenant_uuids):
         self.dissociate(
             line_id, extension_id, tenant_uuid, tenant_uuids, reset_autoprov=True
         )
-        extension_middleware.delete(extension_id, tenant_uuids)
+        try:
+            extension_middleware.delete(extension_id, tenant_uuids)
+        except ResourceError as e:
+            if not str(e).startswith('Resource Error - Extension is associated'):
+                raise e
```
DrPyser, line 44: Instead of matching on exception message, how about creating an exception Subclass(`ExtensionAssociatedError`) and catching only that? I believe that would be less error prone, more concise.

wazo_confd/plugins/user/schema.py
```diff
@@ -369,3 +373,4 @@ class UserLinePutSchema(LineSchema):
 class UserListItemSchemaPut(UserListItemSchema):
     lines = Nested('UserLinePutSchema', many=True)
     agent = Nested('UserAgentSchema', allow_none=True)
+    incalls = Nested('UserIncallPutSchema', many=True)
```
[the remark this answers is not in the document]
DrPyser: This is exposing the extensions in the /user API?

wazo_confd/plugins/extension/validator.py
```diff
@@ -185,6 +185,70 @@ def validate(self, extension):
...
+                    extension_id=extension.id,
+                    associated_id=extension.lines[0].id,
+                )
+            else:
+                raise errors.resource_associated(
+                    'Extension',
+                    'lines',
+                    extension_id=extension.id,
+                    associated_ids=[line.id for line in extension.lines],
+                )
```
DrPyser, line 226: Is there a reason why the distinction between one line and multiple is important?

wazo_confd/plugins/line_extension/middleware.py
xxbillouxx: I totally agree implementing a dedicated exception would be less error prone. Unfortunately I cannot easily create a new Exception, as the ResourceError is created by the validator (wazo_confd/plugins/extension/validator.py). The validators are used when there is a creation, a deletion and an update through the DAOs. The validators are one of the pillars of the DAOs, so there is no easy way to replace/write a new Exception.

wazo_confd/plugins/user/middleware.py
xxbillouxx: see my other comment

wazo_confd/plugins/user/schema.py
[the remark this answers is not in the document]
xxbillouxx: This is not directly related. I found an issue when we try to update a user with their incalls (the call id is ignored). So instead of created a dedicated PR for that and lose time for the other PR to be reviewed, I included this fix in the current PR :-)

wazo_confd/plugins/extension/validator.py
xxbillouxx: Fixed, as there is no specific reason

wazo_confd/plugins/user/schema.py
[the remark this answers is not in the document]
DrPyser: That sounds like a significant issue in the public API though, would be worth describing the issue somewhere(jira) to anticipate reporting, and point to this PR.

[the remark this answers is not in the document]
xxbillouxx: Done :)

wazo_confd/plugins/line_extension/middleware.py
DrPyser: From my understanding, LineExtensionAssociationValidator is only used by the LineExtensionService, so exceptions generated from the validator are only relevant to the service and layers above( API middleware/http resource definition) which use the service. I don't see the DAOs depending on ResourceError.
That being said, I understand not wanting to break the mold here.
But this looks like a bad "dependency" if the xivo_dao lib dictates how exceptions should be used by application-specific code that is not relevant to the DAOs.

DrPyser:
The fix works in preventing portal from deleting an extension associated with another line.
The crux of the issue is in portal though, as the extension is deleted first, and a failure to delete the extension prevents the line from being deleted, but weirdly deassociates the line from the extension.

DrPyser (approved):

DrPyser:
Approved given the scope of the PR, but might not be enough to close WAZO-3221.

wazo_confd/plugins/line/resource.py
```diff
@@ -58,5 +63,5 @@ def put(self, id):
     def delete(self, id):
         tenant_uuid = build_tenant()
         tenant_uuids = self._build_tenant_list({'recurse': True})
-        self._middleware.delete(id, tenant_uuid, tenant_uuids)
+        self._middleware.delete(id, tenant_uuid, tenant_uuids, recursive=is_recursive())
```
DrPyser, line 66: Does the recursive option bypass protection of associated extensions? Will it only delete associated extensions that would be orphaned?

xxbillouxx: It deletes only the extensions if not used at all. I added 2 tests : https://github.com/wazo-platform/wazo-confd/blob/1dab14d800e55f52d18cba2531eb6621626c5e90/integration_tests/suite/base/test_lines.py#L369

DrPyser: Oh thanks!

DrPyser:
Tested recursive delete of lines, everything checks out, good job!

DrPyser:
recheck

xxbillouxx:
recheck

wazo_confd/plugins/line/api.yml
```diff
@@ -84,7 +84,11 @@ paths:
     delete:
       operationId: delete_line
       summary: Delete line
-      description: '**Required ACL:** `confd.lines.{line_id}.delete`'
+      description: '**Required ACL:** `confd.lines.{line_id}.delete`
+
+
+        **Disclaimer**: if `recursive=true`, the line is deleted, all their
```
chartrandf, line 90: Out of curiosity is `recursive` used elsewhere too? We also use `recurse` on GET, could be nice to me consistent on this term.

xxbillouxx: Totally agree with you :) The `recursive` parameter is available on the `/users` endpoint too.
I think `recursive` should be used instead of `recurse` (`recursive` means ... "recursive", this is not the case for `recurse`)

xxbillouxx: I checked the OpenAPI documentation  (like here https://wazo-platform.org/documentation/api/configuration.html#tag/lines/operation/list_lines): `recurse` is used to indicate if the query should include sub-tenants or not.
Here (and in the new `/users` endpoint too) `recursive` is used to indicate that the update/delete/creation should update/delete/create the related resources/associations.
So I think we coud let `recursive` for this PR. What do you think @chartrandf ?

chartrandf: All good, let's keep it like this

chartrandf (approved):

xxbillouxx:
recheck

xxbillouxx:
recheck

PR #391, opened 2023-05-15
add wazo-confd-wait command

fblackburn1:
reason: So that systemd knows when the daemon is ready
It make wazo-acceptance-ha test fail

### DIFF
--- a/wazo_confd/wait.py
+++ b/wazo_confd/wait.py
@@ -1,0 +1,53 @@
+# Copyright 2023 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import socket
+import sys
+import time
+
+from contextlib import closing
+from xivo.chain_map import ChainMap
+from xivo.config_helper import read_config_file_hierarchy
+
+from wazo_confd.config import DEFAULT_CONFIG
+
+HOST = 'localhost'
+TIMEOUT = 60
+INTERVAL = 1
+
+
+def iterations(timeout, interval):
+    end_time = time.time() + timeout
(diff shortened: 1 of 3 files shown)


wazo_confd/wait.py
```diff
@@ -0,0 +1,53 @@
...
+    while time.time() < end_time:
+        yield
+        time_left = end_time - time.time()
+        delay = time_left % interval
+        time.sleep(delay)
+
+
+def tcp_port_is_open(host, port):
+    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as sock:
+        sock.settimeout(INTERVAL)
```
sduthil, line 30: Isn't this redundant with the `time.sleep`? I'm guessing the loop will 1. wait 1 second in the socket timeout and 2. wait 1 second in `time.sleep`, resulting in only about 30 connection tries, instead of the expected 60.

fblackburn1: Well ask to yourself, it's a blindly copy/paste of wazo-auth :smile: 
If you test on a closed port, the socket will return instantaneously, so no duplicate wait
But yes I guess there is a scenario were socket timeout is reached, but it seems not the "normal" behavior
Refactoring improvement can be proposed in https://wazo-dev.atlassian.net/browse/WAZO-1831

sduthil: OK, instant return upon closed port is good for me :)

sduthil (dismissed):

sduthil (approved):

fblackburn1:
recheck

fblackburn1:
recheck (it won't be merged)

PR #392, opened 2023-05-15
WAZO-3256 support commas in pjsip transport option value strings

DrPyser:
## Updated PJSIPSectionOption regex to allow commas in values


pc-m (approved):

DrPyser:
recheck

DrPyser:
recheck

PR #393, opened 2023-05-19
[WAZO-3270] Add logo to spec

chartrandf:
depends-on: https://github.com/wazo-platform/wazo-platform.org/pull/613


jfbourque:
recheck

PR #399, opened 2023-06-26
add a `view` parameters to pass to dao

ja0x:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/217

Why:
* allows to eagerload relationship when needed per route


ja0x:
This fix needs more investigation before being merged

PR #400, opened 2023-06-26
Improve tenant deletion (including some related resources)

xxbillouxx:
Deletion is triggered by the auth_tenant_deleted event and by cron (synd_db.py)

Depends-On: https://github.com/wazo-platform/xivo-dao/pull/225
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/226
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/227

### DIFF
--- a/wazo_confd/plugins/event_handlers/plugin.py
+++ b/wazo_confd/plugins/event_handlers/plugin.py
@@ -1,29 +1,23 @@
-# Copyright 2020-2022 The Wazo Authors  (see the AUTHORS file)
+# Copyright 2020-2023 The Wazo Authors  (see the AUTHORS file)
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
 
-from wazo_provd_client import Client as ProvdClient
-from wazo_provd_client.exceptions import ProvdError
 from xivo_dao import tenant_dao
 from xivo_dao.helpers.db_utils import session_scope
 from xivo_dao.resources.endpoint_sip import dao as sip_dao
 from xivo_dao.resources.pjsip_transport import dao as transport_dao
 
-from wazo_confd.plugins.device.model import Device
-from wazo_confd.plugins.device.notifier import DeviceNotifier
-
 from .service import DefaultSIPTemplateService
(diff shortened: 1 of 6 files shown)


wazo_confd/plugins/event_handlers/plugin.py
```diff
@@ -38,47 +32,16 @@ def _auth_tenant_added(self, event):
...
-
-    def _delete_config(self, config):
-        try:
-            self.provd.configs.delete(config)
-        except ProvdError as e:
-            if e.status_code != 404:
-                raise
-        except Exception:
-            pass  # device has no config
+        remove_tenant(event['uuid'])
```
pc-m, line 35: Where does the provd devices deletion happen now?

wazo_confd/sync_db.py
```diff
@@ -102,10 +102,11 @@ def main():
...
-        user_dao.delete(user)
+    logger.debug('Removing tenant: %s', tenant_uuid)
+    with session_scope():
+        tenant = tenant_resources_dao.get(tenant_uuid)
+        tenant_resources_dao.delete(tenant)
 
-    # FIXME(fblackburn):
-    # * Add all other resources related to tenant_uuid
-    # * Reset device to autoprov
+
```
pc-m, line 110: This is not required, the `console_scripts` in setup.py should automatically add a launcher

wazo_confd/plugins/event_handlers/plugin.py
xxbillouxx: Devices deletion in provd didn't work: the tenant is deleted, so the request to provd is denied (invalid token)

wazo_confd/sync_db.py
xxbillouxx: perfect, but is it a problem if I let that ? (to be able call manually the python script)

wazo_confd/plugins/event_handlers/plugin.py
pc-m: Makes sense

pc-m (approved):

PR #401, opened 2023-07-05
call filters: autogenerate name for multi-tenancy purposes

bloom1:


bloom1:
recheck

bloom1:
recheck

wazo_confd/plugins/call_filter/resource.py
```diff
@@ -14,13 +15,30 @@
...
 
     @required_acl('confd.callfilters.create')
     def post(self):
-        return super().post()
+        form = self.schema().load(request.get_json())
+        form = self.add_tenant_to_form(form)
+
+        tenant = self._tenant_dao.get(form['tenant_uuid'])
+        # NOTE(afournier): we use a UUID as if it was the callfilter UUID but it's not
+        # Call filters do not use UUIDs yet
```
sduthil, line 34: Would it be hard to add the UUID column right now? This would avoid a painful migration (i.e. parsing names to get the UUID) later on when we do add the UUID column.

bloom1: It's another ticket completely. Call filters are far from being the only resource still using an integer ID.
It's also hard, requiring changes in function keys, fallbacks, etc.

integration_tests/suite/base/test_call_filters.py
```diff
@@ -38,11 +38,11 @@ def test_put_errors(call_filter):
 
 
 def error_checks(url):
-    yield s.check_bogus_field_returns_error, url, 'name', 123
-    yield s.check_bogus_field_returns_error, url, 'name', None
-    yield s.check_bogus_field_returns_error, url, 'name', True
-    yield s.check_bogus_field_returns_error, url, 'name', {}
-    yield s.check_bogus_field_returns_error, url, 'name', []
```
sduthil, line 45: Do we have any test covering the deprecated way by sending only a name? We still want to support that use case and it needs at least 2 tests (one for success, one for error/invalid data)

wazo_confd/plugins/call_filter/resource.py
sduthil: And how about adding the column `uuid` but keeping the `id` as primary key, only ensuring that the `uuid` is unique? That way we have the `uuid` in the database, we can use it for the callfilter name and the primary key migration `id` -> `uuid` will be easier.

bloom1:
recheck

PR #423, opened 2023-10-17
context range: add the API for extension listing

pc-m:


integration_tests/suite/base/test_context_range.py
```diff
@@ -0,0 +1,290 @@
...
+                has_entries(
+                    total=2,
+                    items=contains_inanyorder(
+                        {'start': '1000', 'end': '1004'},
+                        {'start': '1006', 'end': '1999'},
+                    ),
+                ),
+            )
+
+            response = confd.contexts(context['id']).ranges('user').get(available=False)
```
xxbillouxx, line 120: IMHO, available=False may mean for people that we want only the extensions that are not available.

Maybe available_only / only_available instead ???

integration_tests/suite/base/test_context_range.py
```diff
@@ -0,0 +1,290 @@
...
+                ),
+            )
+
+    aux()
+
+
+@fixtures.context(
+    user_ranges=[
+        {'start': '00', 'end': '99'},
+        {'start': '0001', 'end': '0499'},
```
xxbillouxx, line 137: Open question: can you confirm we cannot have this case (ranges that conflict each other) :
For example:  
```
{'start': '00', 'end': '99'}
and {'start': '70', 'end': '199'}
```
?

xxbillouxx (commented):
LGTM, thanks a lot.
I have only 2 small comments :-)

integration_tests/suite/base/test_context_range.py
pc-m: You're right, Francis made a similar comment. I'll try to find something better.

integration_tests/suite/base/test_context_range.py
pc-m: Good catch, I'll have to check to be sure that the good thing happens in this case.
Your example does not work since start and end needs to have the same length but
`{'start': '000', 'end': '099'}, {'start': '70', 'end': '199'}` should result in `{'start': '000', 'end': '199'}`

pc-m: The following scenario fails
```python
@fixtures.context(
    user_ranges=[
        {'start': '100', 'end': '119'},
        {'start': '000', 'end': '099'},
        {'start': '070', 'end': '199'},
    ],
)
def test_search_overlapping_ranges(context):
    response = confd.contexts(context['id']).ranges('user').get()
    print(response)
    assert_that(
        response.json,
        has_entries(
            total=1,
            items=contains_inanyorder(
                {'start': '000', 'end': '199'},
            ),
        ),
    )
 ```
 The result with the current implementation is
 ```javascript
 {'items': [{'end': '119', 'start': '000'}], 'total': 1}
 ```

File: wazo_confd/plugins/context_range/api.yml [hunk 3]
```diff
@@ -0,0 +1,73 @@
+paths:
+  /contexts/{context_id}/ranges/{range_type}:
+    get:
+      operationId: list_contexts_range
+      summary: List contexts range
+      description: |
+        **Required ACL:** `confd.contexts.{context_id}.ranges.{range_type}.read`
+
+        List the extension range for a given context and resource type. When filtering
+        using the `available` query string the range will be split into multiple
```
xxbillouxx, line 10: `availability` instead ?

wazo_confd/plugins/context_range/schema.py
```diff
@@ -0,0 +1,18 @@
...
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+from marshmallow import fields
+from marshmallow.validate import Predicate, Length
+from wazo_confd.helpers.mallow import BaseSchema
+from wazo_confd.helpers.restful import ListSchema as BaseListSchema
+
+
+class ListSchema(BaseListSchema):
+    available = fields.Boolean(missing=True, default=True)
```
xxbillouxx, line 11: should be updated ? (availability)

xxbillouxx (commented):
I left 2 comments related to the "availability" param :)

xxbillouxx (approved):
Perfect job, thanks a lot !

PR #425, opened 2023-10-24
user: select query strategy based on pagination

pc-m:
If using large pages (101) or unpaginated queries the existing loading strategy is used. Paginated queries will use the default sqlalchemy behavior, which is lazyloading.

For a small enough number of resource (101 here) this is enough to make the route significantly faster.

Depends-On: https://github.com/wazo-platform/xivo-dao/pull/249


ja0x (approved):

pc-m:
regate

PR #426, opened 2023-10-31
Delete groups and queues when tenant deleted

xxbillouxx:


wazo_confd/sync_db.py
```diff
@@ -127,6 +130,17 @@ def remove_tenant(tenant_uuid, sysconfd):
                 moh.name,
             )
             sysconfd.delete_moh(moh.name)
+
+        logger.debug('Retrieving all queues for tenant: %s', tenant_uuid)
+        queues_list = queue_dao.search(tenant_uuids=[tenant_uuid])
+        for queue in queues_list.items:
+            queue_dao.delete(queue)
```
sduthil, line 137: This is not using foreign keys like other resources. We discussed using a `tenant_uuid` foreign key in the `queue` table. Do you intend to add the foreign key later, or did you have issues with foreign keys?

xxbillouxx: I tried using a fk `tenant_uuid` as you suggested, but I had an issue when updating a `queuefeatures`: the `queue` was automatically updated too, but with `tenant_uuid=None` 😵‍💫 .
I discussed later with Francois: the `queue` table is a very old table (as you said too), it should be removed and its data should be directly integrated into `queuefeatures `and `groupfeatures`. So I choose the fastest/easiest solution now. I created another JIRA ticket https://wazo-dev.atlassian.net/browse/WAZO-3476 to migrate "later" the `queue` table into `queuefeatures` and `groupfeatures`

DrPyser: https://wazo-dev.atlassian.net/browse/WAZO-3408?focusedCommentId=19547

Nabil says in the ticket that `queue` is considered a legacy table that should be integrated into queuefeatures?

PR #435, opened 2023-12-22
Pull request for WAZO-3547-remove-nosetests

sduthil:
## remove conflict between pytest and confd fixtures

How:

* Hack pytest to ignore the x first arguments, replacing the trick used
  for the mock library
* See also:
  https://github.com/pytest-dev/pytest/blob/main/src/_pytest/compat.py#L85

## agent tests: remove agents after creation to avoid conflict

Why:

* Subsequent tests were unable to create agent 1234

## helper bus: avoid early initialization

Why:

* pytest tries to access attributes on all global variables to detect
  which ones are fixtures
* this initializes the bus client too early, before the host and port
  are set

## meeting tests: add missing teardown

Why:

* Other tests will fail because the extension is missing because meetingjoin is disabled

## helpers: avoid early initialization

Why:

* pytest tries to access attributes on all global variables to
  detect which ones are fixtures
* this initializes the clients too early, before the host and
  port are set

## integration test sync db: fix default tenants

Why:

* the created tenant should not exist before the test
* the assertion of empty sip templates for CREATED_TENANT was failing
  with pytest

## test trunk: repair broken tests

Why:

* due to the aux() function, the yielded tests were never executed

## integration tests: replace yields with function calls

Why:

* pytest does not support test generators

## tests: replace nosetests with pytest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@
 .tox
 coverage.xml
 unit-tests.xml
-nosetests.xml
--- a/integration_tests/Makefile
+++ b/integration_tests/Makefile
@@ -26,9 +26,7 @@ build-provd:
 	docker build -t wazoplatform/wazo-provd $(PROVD_DIR)
 
 test:
-	# Explicit match to avoid to include AssetLaunchingTestCase helper methods
-	# This constraint must be removed when tests will be migrated to pytest
-	nosetests --match '^[Tt]est' -x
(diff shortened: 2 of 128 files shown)


sduthil:
recheck

pc-m (approved):
LGTM I don't have time to check the failing test yet

pc-m (approved):

PR #440, opened 2024-01-23
Pull request for WAZO-1870-remove-global-parking

sduthil:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/259

## features: move rejected parking options from dao

Why:

* the DAO does not need it anymore, since the options are removed from
  the database. Those options are now only needed for API validation.

## tests: ignore parking function keys

Why:

* They will be reinstated after the funckeys can use parking lots
  instead of the global parking


fblackburn1:
recheck

fblackburn1 (approved):

fblackburn1:
We have issue on master
But all tests succeed locally
Forcing merge to avoid desync and error 500 with dao

PR #451, opened 2024-03-21
Wp 1205 user incall delete cascade

DrPyser:


sduthil:
This effectively breaks the symmetry between user creation API and user deletion API. IMHO the request should include which resources to delete or not, at least for the incall, agent and voicemail. In some cases, we want to remove them, in other cases we don't.

This is only my opinion, I did not see any discussion about the expected behavior. If you have more info, please share :)

DrPyser:
@sduthil Hmm, i see. The user creation API allows to create lots of resources associated with the user, but many of those resources can also be created separately, associated with the user only temporarily.
If there was a way to know which resources were created along with the user, and which were created separately, it would make sense to have this symmetry.

It's also a matter of perspective, lines and voicemails seems to me like obvious parts of a user, separately they do not make sense(unless voicemails and lines are shared, that is). incalls seem to me more like resources that are independent and can point to destinations variably. I suppose this depends on how incalls are used(is the did pointing to a particular person, or a position in a business which might be filled by different people at different times).

sduthil (approved):

sduthil:
I'm approving the PR because it improves the situation, but IMHO this is not complete. To fix this properly, we need:
* a database migration to cleanup any leftover extension
* API parameters to select which resources should be deleted or unlinked

sduthil (approved):

sduthil:
Selecting which resources will be deleted will be handled in another PR: https://wazo-dev.atlassian.net/browse/WAZO-3688

PR #463, opened 2024-07-11
plugins.ingress_http Allow fallback to master tenant when empty

jirkah:
- added a new fallback query parameter to be able to retrieve a valid ingress from the master tenant if there's no ingress configured on the current tenant

### DIFF
--- a/wazo_confd/helpers/restful.py
+++ b/wazo_confd/helpers/restful.py
@@ -75,10 +75,20 @@ class ListResource(ConfdResource):
         super().__init__()
         self.service = service
 
+    def unsafe_fallback_get(self, fallback_tenants=None):
+        params = self.search_params()
+        tenant_uuids = self._build_tenant_list(params)
+        if fallback_tenants:
+            tenant_uuids.extend(fallback_tenants)
+
+        return self._get(params, tenant_uuids)
+
     def get(self):
         params = self.search_params()
         tenant_uuids = self._build_tenant_list(params)
+        return self._get(params, tenant_uuids)
 
+    def _get(self, params, tenant_uuids):
         kwargs = {}
         if tenant_uuids is not None:
             kwargs['tenant_uuids'] = tenant_uuids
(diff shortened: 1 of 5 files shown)


wazo_confd/plugins/ingress_http/api.yml
```diff
@@ -102,6 +103,13 @@ parameters:
     type: string
     name: http_ingress_uuid
     in: path
+  fallback_to_master:
```
fblackburn1, line 106: We maybe want to be consistent with our logic inside confd and use same logic than `user_external_app` (even if logic is user fallbacking on tenant config)
```
?view=fallback
```

side note: you document `fallback_to_master` but you implement `fallback`

wazo_confd/helpers/restful.py
```diff
@@ -75,10 +75,20 @@ def __init__(self, service):
...
         self.service = service
 
+    def unsafe_fallback_get(self, fallback_tenants=None):
+        params = self.search_params()
+        tenant_uuids = self._build_tenant_list(params)
+        if fallback_tenants:
+            tenant_uuids.extend(fallback_tenants)
+
+        return self._get(params, tenant_uuids)
+
```
fblackburn1, line 85: I would scope this code to ingress plugin until someone else need it (to avoid unnecessary indirection)

wazo_confd/plugins/ingress_http/resource.py
```diff
@@ -28,7 +28,16 @@ def post(self):
 
     @required_acl('confd.ingresses.http.read')
     def get(self):
-        return super().get()
+        result = super().get()
```
fblackburn1, line 31: `Resource` is not really intended to use `super()` when you need a different logic
i.e. The base class has already all helpers to build your logic, so you can just use them explicitly without passing by `super` and the code will be much more readable

wazo_confd/plugins/ingress_http/api.yml
fblackburn1: for the doc - Yes my bad, read too fast

pc-m (approved):

PR #478, opened 2024-09-04
Wazo 3874 sip endpoint list by UUID

pc-m:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/281

### DIFF
--- a/wazo_confd/plugins/api/api.yml
+++ b/wazo_confd/plugins/api/api.yml
@@ -116,6 +116,19 @@ parameters:
     type: string
     description: Search term for filtering a list of items. Only items with a field
       containing the search term will be returned.
+  query_string_uuid_filter:
+    in: query
+    name: uuid
+    required: false
+    collectionFormat: csv
+    type: array
+    minItems: 1
+    maxItems: 20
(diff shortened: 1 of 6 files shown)


ja0x (approved):
Works for me.
Quick question, is it meant to be a "hidden" interface? We don't expose it in the API parameters?

wazo_confd/plugins/api/api.yml
```diff
@@ -116,6 +116,11 @@ parameters:
     type: string
     description: Search term for filtering a list of items. Only items with a field
       containing the search term will be returned.
+  query_string_uuid_filter:
+    name: uuid
+    in: query
+    type: string
+    description: Search filter for resource UUID, can be used to search multiple elements. Using a comma separated list of UUID
```
ja0x, line 123: For the API part, I'd propose adding a few parameters to make the swagger interface more intuitive
```
    required: false
    collectionFormat: csv
    minItems: 1
    maxItems: 20
    uniqueItems: true
    type: array
    items:
      type: string
      format: uuid
```

ja0x (changes requested):

PR #481, opened 2024-09-27
Wazo 3921 numbers api

pc-m:
Depends-On: https://github.com/wazo-platform/wazo-bus/pull/114
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/282
Depends-On: https://github.com/wazo-platform/xivo-manage-db/pull/275


pc-m:
recheck

wazo_confd/plugins/phone_number/notifier.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def edited(self, resource):
+        phone_number_serialized = PhoneNumberSchema().dump(resource)
+        event = PhoneNumberEditedEvent(phone_number_serialized, resource.tenant_uuid)
+        self.bus.queue_event(event)
+
+    def deleted(self, resource):
+        phone_number_serialized = PhoneNumberSchema().dump(resource)
+        event = PhoneNumberDeletedEvent(phone_number_serialized, resource.tenant_uuid)
+        self.bus.queue_event(event)
```
DrPyser, line 32: see wazo-bus comments about support bulking for events.

wazo_confd/plugins/phone_number/schema.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class PhoneNumberSchema(BaseSchema):
+    uuid = fields.UUID(dump_only=True)
+    tenant_uuid = fields.String(dump_only=True)
+    number = fields.String(validate=Length(min=1, max=128), required=True)
+    caller_id_name = fields.String(validate=Length(min=1, max=256), allow_none=True)
+    main = StrictBoolean(default=False)
+    shared = StrictBoolean(default=False)
+    links = ListLink(Link('phone_numbers', field='uuid'))
```
DrPyser, line 17: question: is `rel` attribute of `Link` supposed to be unique/distinguished?

wazo_confd/plugins/phone_number/validator.py
```diff
@@ -0,0 +1,41 @@
...
+        if not phone_number.main:
+            return
+
+        existing = self.dao.find_by(main=True, tenant_uuid=phone_number.tenant_uuid)
+        if existing and existing.uuid != phone_number.uuid:
+            raise errors.resource_exists(
+                'PhoneNumber',
+                main=True,
+                tenant_uuid=str(phone_number.tenant_uuid),
+            )
```
DrPyser, line 28: why do we need to add the overhead of an additional query if we have a database constraint?

pc-m: It's the pattern in wazo-confd where we check first and hope nothing changes after the check. I would prefer handling the IntegrityError but that would be a first in wazo-confd and would need some discussion to introduce this mechanic in the "framework" which has almost no logic other than the marshmallow schema and the sqlalchemy model

integration_tests/suite/helpers/helpers/phone_number.py
```diff
@@ -0,0 +1,28 @@
...
+        response.assert_ok()
+
+
+def add_phone_number(wazo_tenant=None, **parameters):
+    response = confd.phone_numbers.post(parameters, wazo_tenant=wazo_tenant)
+    return response.item
+
+
+def _random_number(length):
+    number = number = ''.join(random.choice(string.digits) for _ in range(length))
```
DrPyser, line 27:
```suggestion
    number = ''.join(random.choice(string.digits) for _ in range(length))
```

wazo_confd/plugins/phone_number/notifier.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .schema import PhoneNumberSchema
+
+
+class PhoneNumberNotifier:
+    def __init__(self, bus):
+        self.bus = bus
+
+    def created(self, resource):
+        phone_number_serialized = PhoneNumberSchema().dump(resource)
```
DrPyser, line 20: I usually/often see a singleton instance defined globally for marshmallow schemas, since most schemas are not stateful. This avoids having to create single use instances.

DrPyser (approved):
LGTM, tested manually.

PR #492, opened 2024-12-16
WAZO-3423 search incalls by user_id

rbienvenu0:
Actual changes were done in the dao, this is only about documentation.

### DIFF
--- a/integration_tests/suite/base/test_incalls.py
+++ b/integration_tests/suite/base/test_incalls.py
@@ -147,6 +147,26 @@ def test_sorting_offset_limit(incall1, incall2):
     s.check_limit(url, incall1, incall2, 'description', 'sort')
 
 
+@fixtures.incall(id=0)
+@fixtures.user(id=0)
+@fixtures.incall(id=1)
+@fixtures.user(id=1)
+def test_list_by_user_id(incall, user1, *_):
+    with a.incall_user(incall, user1):
+        response = confd.incalls().get()
+        assert_that(response.total, 2)
+
+        response = confd.incalls().get(user_id=user1['id'])
+        assert_that(response.total, 1)
+        assert_that(
+            response.items[0],
+            has_entry(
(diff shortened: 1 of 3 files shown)


rbienvenu0:
recheck

bloom1 (approved):
LGTM. Thank you!

integration_tests/suite/base/test_incalls.py
```diff
@@ -165,6 +165,35 @@ def test_get(incall):
     )
 
 
+@fixtures.incall(id=0)
+@fixtures.user(id=1)
+@fixtures.incall(id=1)
+@fixtures.user(id=2)
+def test_get_userid(incall, user1, *_):
```
fblackburn1, line 172: nit: you mix get / list nomenclature for tests. This test should be bellow `test_sorting_offset_limit` and renamed to `test_list_by_user_id`

integration_tests/suite/base/test_incalls.py
```diff
@@ -165,6 +165,35 @@ def test_get(incall):
...
+                has_entries(
+                    {
+                        'moh_uuid': None,
+                        'user_id': user1['id'],
+                        'ring_time': None,
+                        'type': 'user',
+                        'user_firstname': user1['firstname'],
+                        'user_lastname': user1['lastname'],
+                    }
+                ),
```
fblackburn1, line 192:
```suggestion
                has_entries(user_id=user1['id']),
```
This test should focus to only one thing, i.e. that result return the matches user_id

If you want to test if all arguments are there, use `test_get_user_destination_relation` (we can see there is some arguments missing in this test)

integration_tests/suite/base/test_incalls.py
rbienvenu0: Updated

fblackburn1 (approved):

PR #507, opened 2025-03-18
user_blocklist: user api CRUD

DrPyser:
Depends-On: https://github.com/wazo-platform/xivo-dao/pull/302
Depends-On: https://github.com/wazo-platform/xivo-manage-db/pull/301
Depends-On: https://github.com/wazo-platform/wazo-bus/pull/121

### DIFF
--- a/integration_tests/suite/helpers/helpers/blocklist.py
+++ b/integration_tests/suite/helpers/helpers/blocklist.py
@@ -2,13 +2,14 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 import random
-import string
+
+import phonenumbers
 
 from . import confd
 
 
 def generate_user_blocklist_number(confd_client=confd, **parameters):
-    parameters.setdefault('number', _random_number(10))
+    parameters.setdefault('number', _random_number())
     return add_user_blocklist_number(confd_client=confd_client, **parameters)
 
 
@@ -27,6 +28,12 @@ def add_user_blocklist_number(confd_client=confd, wazo_tenant=None, **parameters
     return confd_client.users.me.blocklist.numbers(response.item['uuid']).get().item
 
 
-def _random_number(length):
-    number = ''.join(random.choice(string.digits) for _ in range(length))
-    return f'+{number}'
+def _random_number():
+    random_country_code = random.choice(
+        list(region for region in phonenumbers.SUPPORTED_REGIONS)
+    )
+    number = phonenumbers.example_number_for_type(
+        random_country_code, phonenumbers.PhoneNumberType.FIXED_LINE_OR_MOBILE
+    )
+    assert number
+    return phonenumbers.format_number(number, phonenumbers.PhoneNumberFormat.E164)
--- a/integration_tests/suite/helpers/helpers/blocklist.py
+++ b/integration_tests/suite/helpers/helpers/blocklist.py
@@ -1,0 +1,32 @@
+# Copyright 2025 The Wazo Authors  (see the AUTHORS file)
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+import random
+import string
+
+from . import confd
+
+
+def generate_user_blocklist_number(confd_client=confd, **parameters):
+    parameters.setdefault('number', _random_number(10))
+    return add_user_blocklist_number(confd_client=confd_client, **parameters)
+
+
+def delete_user_blocklist_number(
+    blocklist_number_uuid, confd_client=confd, check=False, **parameters
+):
+    response = confd_client.users.me.blocklist.numbers(blocklist_number_uuid).delete()
+    if check:
+        response.assert_ok()
+
+
+def add_user_blocklist_number(confd_client=confd, wazo_tenant=None, **parameters):
+    response = confd_client.users.me.blocklist.numbers.post(
+        parameters, wazo_tenant=wazo_tenant
+    )
+    return confd_client.users.me.blocklist.numbers(response.item['uuid']).get().item
+
+
+def _random_number(length):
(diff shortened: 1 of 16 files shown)


DrPyser:
recheck

wazo_confd/helpers/common.py
```diff
@@ -42,6 +42,7 @@ def wrapper(*args, **kwargs):
                 logger.error("%s: %s", error, error.details, exc_info=True)
             return [str(error)], error.status_code
         except ProvdError as error:
+            logger.exception(str(error))
```
sduthil, line 45: Are you positive that all `ProvdError` are code errors and should print a Traceback?

wazo_confd/plugins/user_blocklist/plugin.py
```diff
@@ -0,0 +1,31 @@
...
+from wazo_confd.helpers.types import PluginDependencies
+from .resource import (
+    UserMeBlocklistNumberItem,
+    UserMeBlocklistNumberList,
+)
+from .service import build_service
+
+
+class UserBlocklistPluginDependencies(PluginDependencies):
+    pass
```
sduthil, line 13: What is the use of this empty subclass, instead of using `Plugindependencies` directly?

wazo_confd/plugins/user_blocklist/service.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def get_or_create_user_blocklist(self, user_uuid: str) -> Blocklist:
+        try:
+            return self.dao.get_blocklist_by(user_uuid=user_uuid)
+        except NotFoundError:
+            logger.debug('Creating blocklist for user %s', user_uuid)
+            try:
+                user = self.user_dao.get_by(uuid=user_uuid)
+            except Exception:
+                logger.exception('error getting user %s', user_uuid)
```
sduthil, line 40: The error traceback is already logged by `ErrorCatchingResource`, I don't think there's a need for another traceback here. Also, the `user_uuid` is already part of the exception and will be printed in the error message.

So I suggest removing this `try:except` and let the exception be raised and logged automatically

wazo_confd/plugins/user_blocklist/validator.py
```diff
@@ -0,0 +1,33 @@
...
+    def validate(self, model: BlocklistNumber):
+        existing = self.dao.find_by(
+            number=model.number, blocklist_uuid=model.blocklist_uuid
+        )
+        if existing and (not self.update or existing.uuid != model.uuid):
+            raise errors.resource_exists(
+                'BlocklistNumber',
+                number=model.number,
+                blocklist_uuid=model.blocklist_uuid,
+            )
```
sduthil, line 26: This is hard to read, it took me a while to understand the logic. Suggestion:
```suggestion
        if not existing:
            return
        if self.update and existing.uuid == model.uuid:
            return
        raise errors.resource_exists(
                'BlocklistNumber',
                number=model.number,
                blocklist_uuid=model.blocklist_uuid,
            )
```

sduthil (changes requested):

integration_tests/suite/helpers/helpers/blocklist.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    return confd_client.users.me.blocklist.numbers(response.item['uuid']).get().item
+
+
+def _random_number(length):
+    number = ''.join(random.choice(string.digits) for _ in range(length))
+    return f'+{number}'
+
+
+def subsequent_blocklist_number(start_number, skip_count=1):
```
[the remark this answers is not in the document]
DrPyser: nope that's a copy-paste-rename artifact

wazo_confd/helpers/common.py
DrPyser: No, but I think it's better to always log something when there's an unexpected error and then filter out those we don't care about if it's problematic, than fail silently. 
If it was expected, a comment would have been helpful.

wazo_confd/plugins/user_blocklist/plugin.py
DrPyser: Yeah in that case not useful.

wazo_confd/helpers/common.py
DrPyser: I could avoid the traceback and only log the error though.

sduthil (approved):

sduthil (approved):

DrPyser:
> Build failed (gate pipeline). For information on how to proceed, see http://docs.openstack.org/infra/manual/developers.html#automated-testing
> 
> https://zuul.wazo.community/zuul/t/local/buildset/d4b57a61b128422d990c39045acefb42
> 
> ✔️ [tox-linters ](https://zuul.wazo.community/zuul/t/local/build/294a2bbf36bf47fc992a4320b494a218) SUCCESS in 7m 10s ✔️ [wazo-tox-py39 ](https://zuul.wazo.community/zuul/t/local/build/1c4d8aa208f4464e83d4bfd3afd1bca7) SUCCESS in 7m 46s ✔️ [debian-packaging-bullseye ](https://zuul.wazo.community/zuul/t/local/build/a7ab1b33bdcb4e27a950c8bf15872b94) SUCCESS in 3m 45s ❌ [confd-tox-integration ](https://zuul.wazo.community/zuul/t/local/build/35254047798641598666468cde3e970f) **FAILURE** in 22m 47s


> FAILED suite/base/test_extensions.py::test_edit_extension_line_with_exten_outside_range

:shaking_face:

DrPyser:
recheck

DrPyser:
regate

PR #508, opened 2025-03-19
[WAZO-4095] localization: add tenant_uuid to schema

bloom1:
Depends-On: https://github.com/wazo-platform/wazo-bus/pull/119


rbienvenu0 (approved):
LGTM

PR #509, opened 2025-03-21
user import: fix and increase token expiration

fblackburn1:
why: it was 15 min before (comment was wrong) and increase it to 1 hour
when having a big import (>2000 users)
All requests to wazo-auth and wazo-provd slowdown import process ...

edit: for now I'm not sure it's the right solution, since importing 2k users become more and more slow with time
- first users: < 1 second / user
- after 1000 users : > 5 seconds / user

... we may want to improve logic and reduce call to other services (if possible)


fblackburn1:
superset by https://github.com/wazo-platform/wazo-confd/pull/511

PR #511, opened 2025-03-24
Pull request for WAZO-4113-users-import-performance

fblackburn1:
## user import: fix token expiration

why: it was 15 min before (comment was wrong), but 30min is useful when
having a big import (>2000 users)
All requests to wazo-auth and wazo-provd slowdown import process ...

## line_endpoint: remove line validator/notify layer

why: the edit() method is used to call LineFixes helpers.
Using service directly only add an unused layer of validator and send
useless message on bus

The biggest impact is when importing users, the validator will query
wazo-provd for each association, which is ressource consuming

Depends-on: https://github.com/wazo-platform/xivo-dao/pull/303

### DIFF
--- a/etc/nginx/locations/https-available/wazo-confd
+++ b/etc/nginx/locations/https-available/wazo-confd
@@ -4,13 +4,23 @@ location ^~ /api/confd/1.1/guests {
     include /etc/nginx/wazo-no-auth-shared[.]conf;
 }
 
+location ^~ /api/confd/1.1/users/import {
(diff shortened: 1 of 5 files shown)


fblackburn1:
Depends-on: https://github.com/wazo-platform/xivo-dao/pull/303

fblackburn1:
recheck (with depends-on)

fblackburn1:
check experimental

pc-m (approved):

PR #521, opened 2025-05-07
WAZO-4205 Labelled relations

duduclx:
this add label, name or more for more detailled response, so the UI doesn't need extra queries to have a human readable content.

no test have been done.

wdk have not been used.

only edited the source code without any kind of tests.


DrPyser:
Created ticket to track this: https://wazo-dev.atlassian.net/browse/WAZO-4205

wazo_confd/plugins/schedule/schema.py
```diff
@@ -116,15 +116,15 @@ class ScheduleSchema(BaseSchema):
     enabled = fields.Boolean()
     links = ListLink(Link('schedules'))
 
-    incalls = Nested('IncallSchema', only=['id', 'links'], many=True, dump_only=True)
+    incalls = Nested('IncallSchema', only=['id', 'description', 'extensions', 'links'], many=True, dump_only=True)
```
fblackburn1, line 119: I guess the `description` (`extensions` too?) field is here because the `incall` object doesn't have `label`/`name`, so you use `description` field as name.
The problem here is if we merge as it, those fields will become part of the specification and we won't be able to remove them before very long time ...
The proper solution would be to add a new `label` field

fblackburn1 (commented):
Thank you for the contribution :heart: 
I don't have objection to add `label`/`name` for missing object, since this logic is already used for other objects (except for my comment below)
Before merging the PR, we should also add integration tests (ref: `test_get_<object>_relation`, example: [`test_get_extension_relation`](https://github.com/wazo-platform/wazo-confd/blob/master/integration_tests/suite/base/test_group_extension.py#L220-L235)). 
Also update documentation (`api.yml`)
disclaimer: I didn't test the PR :sweat_smile:

PR #562, opened 2026-03-31
[BUG-32] [WP-1719] Expose caller ID name in outgoing caller ids 

bloom1:
> [!NOTE]
> **Low Risk**
> Low risk additive API change: extends `GET /1.1/users/{uuid}/callerids/outgoing` responses with a new read-only `caller_id_name` field and updates tests/docs accordingly.
> 
> **Overview**
> Adds a read-only `caller_id_name` to `GET /1.1/users/{uuid}/callerids/outgoing` items, populating it for `main` and `shared` caller IDs from the underlying phone-number record (defaulting to an empty string when unset).
> 
> Updates the OpenAPI schema, marshmallow serialization (omitting the new field for `anonymous` entries), changelog, and integration/unit tests to validate the new response shape.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0a1a3064327244ac151618a1b56c97639368445d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/wazo_confd/plugins/user_callerid/service.py
+++ b/wazo_confd/plugins/user_callerid/service.py
@@ -15,6 +15,7 @@ from .types import CallerIDType
 class CallerID:
     type: CallerIDType
     number: str = ''
+    caller_id_name: str = ''
 
 
 CallerIDAnonymous = CallerID(type='anonymous')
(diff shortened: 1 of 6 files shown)


wazo_confd/plugins/user_callerid/service.py
```diff
@@ -42,20 +43,30 @@ def search(self, user_id, tenant_uuid, parameters):
...
+                    number=main_callerid.number,
+                    caller_id_name=main_callerid.caller_id_name or '',
+                )
+            )
 
         # consider "associated" caller ids from incalls
         # as having precedence over shared phone numbers
         callerids.extend(
-            callerid
+            CallerID(type='associated', number=callerid.number)
```
rbienvenu0, line 57: Shouldn't `caller_id_name` be added here too?

bloom1: That would have been nice indeed, but this is the associated caller ids that come from incalls, which do not have caller id configuration for when they are used as outgoing calls. There are two fields in the incall schema that makes it seem like it's possible (exactly `caller_id_name` and `caller_id_mode`), but that is used to replace the incoming caller ID using the mode (prepend, append or replace). So we do not want to use that `caller_id_name`, because it's a name used to facilitate the knowledge of the source of an incall.
For example, someone might want to see "Sales - \<real caller ID\>" as the caller ID for an incall, so they would configure the `caller_id_mode` to `prepend` and `caller_id_name` to `Sales`.

rbienvenu0 (approved):

PR #563, opened 2026-04-09
[ITEM-245] trigger res_pjsip reload after deleting trunk

bloom1:
> [!NOTE]
> **Medium Risk**
> Adds PBX reload side-effects to tenant-removal cleanup and asserts the behavior in integration tests, which could impact production reload frequency/timing during bulk deletions. Logic is straightforward but touches teardown of tenant resources and sysconfd handler execution.
> 
> **Overview**
> When a tenant is removed via `wazo_confd/sync_db.py`, trunk cleanup is now included: all trunks for the tenant are deleted and `sysconfd` reload handlers are queued (`module reload res_pjsip.so` for SIP trunks, `iax2 reload` for IAX trunks).
> 
> Integration tests add coverage ensuring tenant deletion triggers a PJSIP reload when a SIP trunk exists, and `test_live_reload_false` now resets live reload to `True` to avoid making `sysconfd` assertions a no-op.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fb5156aff9117464e24e9904dc815d28d8953f4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/integration_tests/suite/base/test_configuration.py
+++ b/integration_tests/suite/base/test_configuration.py
@@ -39,6 +39,9 @@ def test_live_reload_false():
     response = confd.configuration.live_reload.get()
     assert_that(response.item, expected)
 
+    # Reset to True so tests using sysconfd are not noop
+    confd.configuration.live_reload.put({'enabled': True})
+
 
 def test_restrict_only_master_tenant():
     response = confd.configuration.live_reload.get(token=TOKEN_SUB_TENANT)
--- a/integration_tests/suite/base/test_event_handlers.py
+++ b/integration_tests/suite/base/test_event_handlers.py
@@ -68,7 +68,7 @@ def test_delete_tenant_reloads_pjsip_for_trunks(trunk, sip):
             response = confd.trunks(trunk['id']).get(wazo_tenant=DELETED_TENANT)
             response.assert_status(404)
 
-        until.assert_(pjsip_reloaded_and_trunk_deleted, tries=5, interval=5)
+        until.assert_(pjsip_reloaded_and_trunk_deleted, tries=5)
 
 
 @pytest.mark.skip(reason="Too flaky because of WAZO-2752")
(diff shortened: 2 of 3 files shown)


bloom1:
recheck

integration_tests/suite/base/test_event_handlers.py
```diff
@@ -50,6 +50,27 @@ def slug_created():
...
+            sysconfd.assert_request(
+                '/exec_request_handlers',
+                method='POST',
+                json={'ipbx': ['module reload res_pjsip.so']},
+            )
+
+            response = confd.trunks(trunk['id']).get(wazo_tenant=DELETED_TENANT)
+            response.assert_status(404)
+
+        until.assert_(pjsip_reloaded_and_trunk_deleted, tries=5, interval=5)
```
fblackburn1: interval 5 is huge, leave default or explicit set the same value
If we need to wait more than 5 second (i.e. 5 tries), then there is an issue somewhere else
```suggestion
        until.assert_(pjsip_reloaded_and_trunk_deleted, tries=5, interval=1)
```

integration_tests/suite/base/test_configuration.py
```diff
@@ -39,6 +39,9 @@ def test_live_reload_false():
     response = confd.configuration.live_reload.get()
     assert_that(response.item, expected)
 
+    # Reset to True so tests using sysconfd are not noop
+    confd.configuration.live_reload.put({'enabled': True})
```
fblackburn1, line 43: 🤔 does that mean other tests using sysconfd (ha, dhcp, voicemail) are false negative?

bloom1: Maybe, but the live_reload config stays persistent across tests so we have to reset it correctly.

fblackburn1: After discussion, they don't use request_handler, but use sysconfd directly

fblackburn1 (approved):

PR #573, opened 2026-09-16
ITEM-700: persist a default outgoing caller ID per user

bloom1:
Depends-On: wazo-platform/wazo-confd-client#60

Why: outgoing caller ID selection is app-only. The choice lives in browser local
storage and is asserted per call via the `X-Wazo-Selected-Caller-ID` SIP header.
A Yealink desk phone cannot send that header, so the same user presents a
different number depending on the device.

wazo-agid already falls back to `userfeatures.outcallerid` when no header
arrives, which is exactly the hardphone case. What was missing was a safe way to
set that value: the only writer was `PUT /users/{id}`, behind the tenant-wide
`confd.users.{id}.update` ACL, and the field had no validation at all.

## Summary

Three commits, reviewable independently.

**`feat(user-callerid)`** adds `GET|PUT /1.1/users/{user_id}/callerids/outgoing/default`
and the `/users/me` equivalents, plus `GET /1.1/users/me/callerids/outgoing`.

An update is accepted only when the number is one the user may already present,
resolved against the existing available-caller-ID list with the same
`phonenumbers` semantic match the list builder uses. Anything else is rejected
with a 400 naming the acceptable numbers. Accepted numbers are normalized to
`+E.164` using the tenant country and stored as `"Name" <+E164>`, the form the
dialplan parses back out.

Reading reports `custom` when the stored caller ID is no longer one available to
the user, for example after the phone number is removed from the tenant, rather
than pretending the value is still usable.

No new column, table or migration: `userfeatures.outcallerid` already exists and
is already read by the AGI on every outcall. Writes emit `UserEditedEvent` but
ask sysconfd for nothing, since wazo-agid reads the column per call and an end
user may change this often.

**`fix(user)`** validates `outgoing_caller_id`, which previously accepted any
string up to 80 characters.

**`docs(user)`** corrects that field's contract in `api.yml`. It was documented
as an enum of `default`, `anonymous` and `'"Firstname Lastname" < number >`.
That is not what the dialplan accepts: `CALLERID_MATCHER` also takes an unquoted
name, a bare number, and a name on its own, all of which are in use. An existing
integration fixture stores `Hûrsule <4441234567>`, which the enum would reject.
The pattern is copied from the schema module so the two cannot disagree.

## Tests

- 2080 passed, 24 skipped in the confd integration suite, unchanged from the
  baseline before these commits
- 601 unit tests, all pre-commit hooks clean
- 12 new integration tests in `test_user_callerid.py` covering the round trip,
  the 400 cases, tenant isolation, per-user isolation, and the `custom` path
- `test_documentation.py` passes, so the served spec still validates

## Not covered

Nothing here has run against a live telephony stack. That a real handset
presents the chosen number to a real callee, with the trunk caller ID format
applied on the wire, is still unverified.

## Related

- wazo-agid applies the trunk caller ID format to stored values, without which
  this feature presents unformatted numbers on the wire. No CI dependency, but
  the two belong in the same release.
- wazo-auth-keys grants end users the ACL for the `/users/me` write.
- wazo-confd-client carries the client methods for these endpoints, and the
  confd test job needs it, hence the `Depends-On` above.
- the uc-doc page documents the new endpoints.

---

> [!NOTE]
> **Medium Risk**
> Changes telephony-facing caller ID storage and adds stricter validation on `outgoing_caller_id`, which may reject updates that were previously accepted; new default endpoints enforce allowed numbers, reducing arbitrary caller ID assertion risk.
> 
> **Overview**
> Adds **GET/PUT default outgoing caller ID** APIs for admins (`/users/{id}/callerids/outgoing/default`) and end users (`/users/me/callerids/outgoing` + `.../default`), persisting to existing `userfeatures.outcallerid` without a migration. Updates are **restricted to numbers the user can already present** (from the outgoing caller ID list), normalized to **+E.164** when possible; unavailable stored values are read back as type **`custom`**.
> 
> **Tightens `outgoing_caller_id` on `PUT /users`**: replaces the inaccurate OpenAPI enum with a dialplan-aligned pattern (wazo-agid `CALLERID_MATCHER`), returning **400** for values the dialplan cannot parse (e.g. bare `<number>`, bad quotes, `;`/`@` in names).
> 
> Writes on the default endpoint emit **`UserEditedEvent`** only (no sysconfd/dialplan reload), since AGI reads the column per call.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b2d8bd6b87e38049e931bb8dd781e5334a04c965. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bloom1:
Confirmed, and it was worse than reported: the write did not just fail, it
returned a 500 leaking the failing SQL.

```
500 ["Unexpected error: (psycopg2.errors.StringDataRightTruncation)
     value too long for type character varying(80)
     [SQL: UPDATE userfeatures SET outcallerid=%(outcallerid)s ...
```

Reproduced with a shared phone number carrying a 256 character
`caller_id_name`, which is the maximum the phone number schema allows, then
selecting it as the default.

There is a second path the report did not mention: `PhoneNumber()` allows 128
characters, so a long enough number overflows the same 80 character column on
its own, with no name involved.

Fixed in b7e8214f. The name is cosmetic and the number is the point, so the name
is dropped when the pair will not fit; a number that alone exceeds the column is
rejected with a 400. The limit is read from `UserFeatures.outcallerid.type.length`
rather than written down, so it cannot drift from the column.

Covered by three unit tests and two integration tests, including the exact
reproduction above.

bloom1:
Partly valid. The location and the cause are right, the symptom is not.

`parse_caller_id` does only understand the quoted form, but `same_phone_number`
is lenient enough that the lookup still succeeds, so the type is resolved
correctly and `custom` is not reported. Reproduced against a running stack, with
a shared number `+14445551234` available to the user:

```
stored 'Hursule <4445551234>'   -> {'type': 'shared', 'number': 'Hursule <4445551234>', 'caller_id_name': 'Acme'}
stored '"Acme" <+14445551234>'  -> {'type': 'shared', 'number': '+14445551234',         'caller_id_name': 'Acme'}
stored 'Acme Corp'              -> {'type': 'custom', 'number': 'Acme Corp',            'caller_id_name': ''}
```

The real defect is on the first line: `number` carries the whole stored string.
That field is documented and typed as a phone number, so a client reading the
default and writing it back is rejected by the `PhoneNumber` validation, which
makes the value not round-trippable. The third line shows the same shape for a
name with no number.

Fixed in b2d8bd6b by parsing the shapes wazo-agid's `CALLERID_MATCHER` accepts:
quoted or unquoted name, optional number, and a name that looks like a number
treated as one. Covered by three unit tests and one integration test that sets
the unquoted form through `PUT /users/{id}` and asserts the number reads back as
a number.

